PR #6, opened 2023-10-22
chore: check output public signals

### LINKED ISSUE #2: Ensure output is checked for Keccak
# Problem

In Keccak example https://github.com/oskarth/mopro/blob/main/mopro-core/src/middleware/circom/mod.rs#L228 we are currently not verifying that the output is as expected, only that the proof is verified.

## Notes

This test vector should be used when verifying: https://github.com/oskarth/mopro/blob/main/mopro-core/src/middleware/circom/mod.rs#L250-L254 (From https://github.com/vocdoni/keccak256-circom/tree/master/test/circuits)

See circom-compat for examples.

This should be exposed through mopro-core, mopro-ffi and iOS app as appropriate.

## Acceptance criteria

Output checked and verified. Able to see and verify from iOS app.

vivianjeng:
Check public signals with `assert_eq!(serialized_inputs, serialized_outputs);`
close #2


oskarth (commented):
Thanks! Could we also check this in the other parts of the code base?

- mopro-ffi https://github.com/oskarth/mopro/blob/main/mopro-ffi/src/lib.rs#L198
- https://github.com/oskarth/mopro/blob/main/mopro-ffi/tests/bindings/test_mopro_keccak.swift#L39
- https://github.com/oskarth/mopro/blob/main/mopro-ios/MoproKit/Example/MoproKit/ViewController.swift#L89

It'd be useful to verify output end to end as this is often needed for application logic, and it'd be a good example of how to do this

mopro-ffi/Cargo.toml
```diff
@@ -33,3 +33,4 @@ uniffi = { version = "0.24", features = ["build"] }
 
 [dev-dependencies]
 uniffi = { version = "0.24", features = ["bindgen-tests"] }
+ark-bn254 = { version = "=0.4.0" }
```
oskarth, line 36: Newline

mopro-ffi/tests/bindings/test_mopro_keccak.swift
```diff
@@ -18,6 +18,31 @@ func bytesToBits(bytes: [UInt8]) -> [Int32] {
     return bits
 }
 
+// TODO: should handle 254-bit input
+func serializeOutputs(_ int32Array: [Int32]) -> [UInt8] {
```
oskarth, line 22: If this is a useful utility that is annoying to code in Swift, we could consider exposing it through FFI. What do you think? Maybe not in this PR but in general as part of "dealing with serialization stuff" https://github.com/oskarth/mopro/blob/main/mopro-core/src/middleware/circom/serialization.rs

mopro-ios/MoproKit/Example/MoproKit/ViewController.swift
```diff
@@ -160,3 +171,28 @@ func bytesToBits(bytes: [UInt8]) -> [Int32] {
...
+            byteLength = byteLength + $0.count
+        }
+        if byteLength < targetLength {
+            let paddingCount = targetLength - byteLength
+            let paddingArray = [UInt8](repeating: 0, count: paddingCount)
+            bytesArray.append(contentsOf: paddingArray)
+        } 
+    }
+    return bytesArray
+}
```
oskarth, line 198: Newline

oskarth (approved):
Nice one! Just a comment re serialization (can do separately) and two minor tweaks. LGTM

mopro-ios/MoproKit/Example/MoproKit/ViewController.swift
```diff
@@ -160,3 +171,28 @@ func bytesToBits(bytes: [UInt8]) -> [Int32] {
     }
     return bits
 }
+
+// TODO: should handle 254-bit input
+func serializeOutputs(_ int32Array: [Int32]) -> [UInt8] {
```
oskarth, line 176: Actually why do we have to do serializeOutputs here when we don't do this for inputs? Shouldn't it be the same?

PR #7, opened 2023-10-23
Spike/real device

oskarth:
After more than 30h and over more than a week, with lots of attempts and many different changes I think I got real device to work.

See https://github.com/oskarth/mopro/issues/1

I'll factor out more minimal diffs from this to make it more usable. The key diff seems to be using a wasmer memory patch.

This PR is also sketching out dylib use, which might not be 100% needed, but is something we might want down the line.

Screenshot from real device

image


oskarth:
Minimal diff for wasmer only https://github.com/oskarth/mopro/pull/8

I'll separate out the dylib stuff at some point as I think we want it

oskarth:
Rust dylib and wasmer stuff integrated. The PR still has some Xcode/script updates that could be useful, so leaving it open for now.

oskarth:
Out of date with recent work

PR #24, opened 2023-10-30
fix(core, ffi): should accept bigint inputs

vivianjeng:
- **Serialization:** `Fr` -> `BigUInt` -> `String`
- **Deserialization:** `String` -> `BigUInt` -> `Fr`
- Add tests for scalar field input: `21888242871839275222246405745257275088548364400416034343698204186575808495616`

### DIFF
--- a/mopro-ffi/src/lib.rs
+++ b/mopro-ffi/src/lib.rs
@@ -4,6 +4,7 @@ use mopro_core::MoproError;
 use num_bigint::BigInt;
 
 use std::collections::HashMap;
+use std::str::FromStr;
 use std::sync::RwLock;
 
 #[derive(Debug)]
@@ -65,14 +66,21 @@ impl MoproCircom {
 
     pub fn generate_proof(
         &self,
-        inputs: HashMap<String, Vec<i32>>,
+        inputs: HashMap<String, Vec<String>>,
     ) -> Result<GenerateProofResult, MoproError> {
         let mut state_guard = self.state.write().unwrap();
 
         // Convert inputs to BigInt
         let bigint_inputs = inputs
             .into_iter()
-            .map(|(k, v)| (k, v.into_iter().map(|i| BigInt::from(i)).collect()))
+            .map(|(k, v)| {
+                (
+                    k,
+                    v.into_iter()
+                        .map(|i| BigInt::from_str(&i).unwrap())
+                        .collect(),
+                )
+            })
             .collect();
 
         let (proof, inputs) = state_guard.generate_proof(bigint_inputs)?;
@@ -119,10 +127,14 @@ uniffi::include_scaffolding!("mopro");
 mod tests {
     use super::*;
     use ark_bn254::Fr;
+    use num_bigint::BigUint;
 
-    fn bytes_to_circuit_inputs(input_vec: &Vec<u8>) -> HashMap<String, Vec<i32>> {
+    fn bytes_to_circuit_inputs(input_vec: &Vec<u8>) -> HashMap<String, Vec<String>> {
         let bits = circom::utils::bytes_to_bits(&input_vec);
-        let converted_vec: Vec<i32> = bits.into_iter().map(|bit| bit as i32).collect();
+        let converted_vec: Vec<String> = bits
+            .into_iter()
+            .map(|bit| (bit as i32).to_string())
+            .collect();
         let mut inputs = HashMap::new();
         inputs.insert("in".to_string(), converted_vec);
         inputs
@@ -156,11 +168,14 @@ mod tests {
         assert!(setup_result.provingKey.len() > 0);
 
         let mut inputs = HashMap::new();
-        let a = 3;
-        let b = 5;
-        let c = a * b;
-        inputs.insert("a".to_string(), vec![a]);
-        inputs.insert("b".to_string(), vec![b]);
+        let a = BigUint::from_str(
+            "21888242871839275222246405745257275088548364400416034343698204186575808495616",
+        )
+        .unwrap();
+        let b = BigUint::from(1u8);
+        let c = a.clone() * b.clone();
+        inputs.insert("a".to_string(), vec![a.to_string()]);
+        inputs.insert("b".to_string(), vec![b.to_string()]);
         // output = [public output c, public input a]
         let expected_output = vec![Fr::from(c), Fr::from(a)];
         let circom_outputs = circom::serialization::SerializableInputs(expected_output);
--- a/mopro-ffi/src/mopro.udl
+++ b/mopro-ffi/src/mopro.udl
@@ -27,7 +27,7 @@ interface MoproCircom {
   SetupResult setup(string wasm_path, string r1cs_path);
 
   [Throws=MoproError]
-  GenerateProofResult generate_proof(record<string, sequence<i32>> circuit_inputs);
+  GenerateProofResult generate_proof(record<string, sequence<string>> circuit_inputs);
 
   [Throws=MoproError]
   boolean verify_proof(bytes proof, bytes public_input);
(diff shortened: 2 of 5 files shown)


mopro-core/src/middleware/circom/mod.rs
```diff
@@ -290,14 +293,14 @@ mod tests {
...
+        let field = "21888242871839275222246405745257275088548364400416034343698204186575808495616";
+        let a = field;
+        let b = "1";
+        let c = field;
+        inputs.insert("a".to_string(), vec![BigInt::from_str(a).unwrap()]);
+        inputs.insert("b".to_string(), vec![BigInt::from_str(b).unwrap()]);
         // output = [public output c, public input a]
-        let expected_output = vec![Fr::from(c), Fr::from(a)];
-        let serialized_outputs = SerializableInputs(expected_output);
+        let serialized_outputs = deserialize_inputs(vec![c.to_string(), a.to_string()]);
```
oskarth, line 303: Why this change? Shouldn't it still be something Serializable?

mopro-core/src/middleware/circom/serialization.rs
```diff
@@ -37,16 +39,22 @@ pub fn deserialize_proving_key(data: Vec<u8>) -> SerializableProvingKey {
         .expect("Deserialization failed")
 }
 
-pub fn serialize_inputs(inputs: &SerializableInputs) -> Vec<u8> {
-    let mut serialized_data = Vec::new();
-    inputs
-        .serialize_uncompressed(&mut serialized_data)
-        .expect("Serialization failed");
+pub fn serialize_inputs(inputs: &SerializableInputs) -> Vec<String> {
+    let serialized_bigint_data: Vec<BigUint> =
```
oskarth, line 43: I wonder if this function can be written more in same style as old one? Not sure diff has to be this big with this many operations (could be wrong though)

mopro-core/src/middleware/circom/serialization.rs
```diff
@@ -37,16 +39,22 @@ pub fn deserialize_proving_key(data: Vec<u8>) -> SerializableProvingKey {
...
+        .into_iter()
+        .map(|x| x.to_string())
+        .collect();
     serialized_data
 }
 
-pub fn deserialize_inputs(data: Vec<u8>) -> SerializableInputs {
-    SerializableInputs::deserialize_uncompressed(&mut &data[..]).expect("Deserialization failed")
+pub fn deserialize_inputs(data: Vec<String>) -> SerializableInputs {
+    let deserialized_data = data
```
oskarth, line 53: Ditto this - can we re-use existing serialization methods? don't see anything immediate that'd disqualify strings

Also think for both functions we can actually just add another function because signature is different, so don't need to "overwrite" existing ones in case we want it for other fields

mopro-core/src/middleware/circom/utils.rs
```diff
@@ -21,12 +21,12 @@ pub fn assert_paths_exists(wasm_path: &str, r1cs_path: &str) -> Result<(), Mopro
     Ok(())
 }
 
-pub fn bytes_to_bits(bytes: &[u8]) -> Vec<bool> {
```
oskarth, line 24: Why his change?

mopro-ffi/src/lib.rs
```diff
@@ -65,14 +66,21 @@ impl MoproCircom {
 
     pub fn generate_proof(
         &self,
-        inputs: HashMap<String, Vec<i32>>,
+        inputs: HashMap<String, Vec<String>>,
```
oskarth, line 69: This makes sense to me

mopro-ffi/src/lib.rs
```diff
@@ -156,15 +165,13 @@ mod tests {
...
-        let expected_output = vec![Fr::from(c), Fr::from(a)];
-        let circom_outputs = circom::serialization::SerializableInputs(expected_output);
-        let serialized_outputs = circom::serialization::serialize_inputs(&circom_outputs);
+        let field = "21888242871839275222246405745257275088548364400416034343698204186575808495616";
+        let a = field.to_string();
+        let b = "1".to_string();
+        let c = field.to_string();
+        inputs.insert("a".to_string(), vec![a.clone()]);
+        inputs.insert("b".to_string(), vec![b.clone()]);
+        let serialized_outputs = vec![c, a];
```
oskarth, line 174: I'd expect this to call serialization method

mopro-ffi/src/mopro.udl
```diff
@@ -12,7 +12,7 @@ dictionary SetupResult {
 
 dictionary GenerateProofResult {
   bytes proof;
-  bytes inputs;
+  sequence<string> inputs;
```
oskarth, line 15: Do we need this change? If it is a byte string we can just deserialize it higher up?

It might make sense, just seems orthogonal to accepting bigint inputs

mopro-ffi/tests/bindings/test_mopro.swift
```diff
@@ -6,51 +6,26 @@ let moproCircom = MoproCircom()
...
     assert(!setupResult.provingKey.isEmpty, "Proving key should not be empty")
 
     // Prepare inputs
-    var inputs = [String: [Int32]]()
-    inputs["a"] = [3]
-    inputs["b"] = [5]
+    var inputs = [String: [String]]()
+    let field = "21888242871839275222246405745257275088548364400416034343698204186575808495616"
+    inputs["a"] = [field]
+    inputs["b"] = ["1"]
```
oskarth, line 18: In Swift, perhaps you'd still do normal math operations on numbers, and then only do the serialization to string at the very end before calling the FFI?

mopro-ffi/tests/bindings/test_mopro.swift
```diff
@@ -6,51 +6,26 @@ let moproCircom = MoproCircom()
 let wasmPath = "./../../../../mopro-core/examples/circom/multiplier2/target/multiplier2_js/multiplier2.wasm"
 let r1csPath = "./../../../../mopro-core/examples/circom/multiplier2/target/multiplier2.r1cs"
 
-// TODO: should handle 254-bit input
-func serializeOutputs(_ int32Array: [Int32]) -> [UInt8] {
```
oskarth, line 10: I imagine we have functions like

`func serializeInputs(_ int32Array: [Int32]) -> [String]`

// TODO: Add proper BigNumber library
// For now can mock with `struct BigNumber` and then `BigNumber("1532423423432")`

`func serializeInputs(_ BNArray: [BigNumber]) -> [UInt8]`

Since Outputs and Inputs are symmetrical, and there's really no "output", perhaps we can just call this serializeInputs? If we think it is confusing to not have a serializeOutputs method we can just have it be identical / alias to serializeInputs.

Then similarly for deserialization

`func deserializeInputs(_ strArray: [String]) -> [BigNumber]`

That way the fact that we serialize to strings is an implementation detail (this could change in the future) and we can easily add serialization methods for various types

Use of BigNumber in Swift seems like a separate issue and more of an app concern for now, so just mocking this with a basic type wrapping string in app/tests seems OK for now

mopro-core/src/middleware/circom/mod.rs
vivianjeng: Just to simplify the serialization
or we will need
```rust
Fr::from(BigUint::from_str(a).unwrap()))
```
to construct the element from string

mopro-ffi/tests/bindings/test_mopro.swift
vivianjeng: I think we can compute over BigNumber before calling `generateProof`
(there should be a `toString` like function in such library)
but if we can handle `serialize_inputs` and `deserialize_inputs` in rust it could be more easier for swift developers?
I prefer not to have these two functions `serializeInputs` and `deserializeInputs` in swift

oskarth (approved):

PR #26, opened 2023-11-03
Feat/ark zkey

oskarth:
Experimenting with different ways of loading zkey


oskarth:
Pushed ark-zkey crate to main instead. Can integrate once fixed performance

PR #35, opened 2023-11-11
feat(core): Add Anon Aadhaar proof generation view

Meyanis95:
This add a new view to generate Aadhaar proofs using RSA verify circuits


mopro-ios/MoproKit/Example/MoproKit.xcodeproj/project.pbxproj
```diff
@@ -253,6 +256,7 @@
 				TargetAttributes = {
 					607FACCF1AFB9204008FA782 = {
 						CreatedOnToolsVersion = 6.3.1;
+						DevelopmentTeam = 7BMWCJP4MA;
```
oskarth, line 259: Ideally we can remove this, but should be OK to merge without

oskarth (approved):
Awesome! Will go ahead and merge, if we need any cleanup tweaks can do after

PR #37, opened 2023-11-12
WIP: feat(ios): Use RSA as dylib

oskarth:
Fix standalone device problem.

Needs to be cleaned up to be nicer, but works on its own now.

Don't merge with .dylib assets. These should be copied over appropriately etc.


oskarth:
Out of date with recent work

PR #38, opened 2023-11-12
Feat/adhaar dylib

oskarth:
Aadhaar with dylib. Should work without being plugged in to laptop.

Don't merge this; dylib hardcoded, not par of build process.

(part of zkhack https://devfolio.co/projects/two-and-a-half-zk-pocketaadhaar-and-pocketkimchi-a8a6)


oskarth:
Out of date with recent work

PR #48, opened 2023-12-07
ci: Improve CI performance

oskarth:
Try to improve CI speed with parallelism, matrix builds and caching

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -6,11 +6,12 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  # TODO: Add a job to run clippy
   lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Check mopro-core formatting
+      - name: Check ark-zkey formatting
         run: cd ark-zkey/ && cargo fmt --all -- --check
       - name: Check mopro-core formatting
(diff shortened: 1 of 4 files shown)


vivianjeng:
we can also remove install circom, since in prepare CI we download all keys
and lint can be run parallelly 

lint:
    ~~needs: test~~

oskarth:
Setup can probably be made faster, especially on Macos https://github.com/oskarth/mopro/actions/runs/7128695177/job/19411159865

Ideally we can save the files as opposed to downloading them from different data center.

Android build is failing, not sure why - could be related to how the matrix build is operating. Assumed this would take the ubuntu prepare step env variables

oskarth:
```
  = note: the `aarch64-linux-android` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-linux-android`
  ```

Confused how this worked before :D

**Comment by vivianjeng** [comment]
Actually the iOS build also fails but it didn't throw error

截圖 2023-12-07 晚上8 31 17

maybe you can check if the file (`wasm`, `arkzkey`) exists (should read from cache?)
the error says
```
error: couldn't read /home/runner/work/mopro/mopro/mopro-core/examples/circom/keccak256/target/keccak256_256_test_js/keccak256_256_test.wasm: No such file or directory (os error 2)
```

oskarth:
Superseeded by https://github.com/oskarth/mopro/pull/51w which is a simpler incremental improvement, less error prone

Issue #21, opened 2023-10-28
Implement micro benchmarking tooling for doing MSMs/using GPU on Mobile


oskarth:
Part of https://github.com/oskarth/mopro/issues/22

## Problem

We currently lack basic tooling for doing micro benchmarking on mobile.

## Steps

1. Create `gpu_explorations` folder in [`mopro-core`](https://github.com/oskarth/mopro/tree/main/mopro-core/src/middleware) with `run_msm_bench()` function that runs N MSM, with timings
2. Add and use vanilla [arkworks msm](https://github.com/arkworks-rs/algebra/blob/master/ec/src/scalar_mul/fixed_base.rs#L85)
3. Extend FFI in [lib.rs](https://github.com/oskarth/mopro/blob/main/mopro-ffi/src/lib.rs) and UDL file to support above function
4. Write basic [Swift test](https://github.com/oskarth/mopro/tree/main/mopro-ffi/tests/bindings) to check initial performance on laptop (not mobile/iOS)
5. Add and call above function in the iOS example app [ViewController.swift](https://github.com/oskarth/mopro/blob/main/mopro-ios/MoproKit/Example/MoproKit/ViewController.swift)
6. Run on a real iOS device to check performance
7. Update README with instructions for how to run above; add initial benchmarks observed

## Next steps

After basics are in place, we can do things like:
- Add alternative implementations like https://lib.rs/crates/ark-msm / Zprize work
- Add feature flags for specific dependencies/versions
- Get it to run on iOS GPU
- Do mobile specific optimizations
- Integrate into actual proof systems, etc

## Acceptance Criteria

Basic micro benchmarking tooling is in place with initial numbers running on a real iOS device.

oskarth:
@FoodChain1028 @moven0831 can this be closed with https://github.com/oskarth/mopro/pull/71?

Would you like to turn next steps into specific GH issues with acceptance criteria and relevant notes in this repo?

moven0831:
The issue is closed by #71.
The next steps of MSM acceleration with mobile GPU will be published in another issue ASAP.

ssxssx:
Can we still apply for the open task in the https://github.com/privacy-scaling-explorations/acceleration-program/issues/24? May I ask whether this project and task have been finished, and are they no longer open in the acceleration-program?

NOOMA-42:
the original acc program issue points at this issue which is wrong, should be https://github.com/oskarth/mopro/issues/15
[...]

PR #53, opened 2024-01-12
feat(benchmarks): msm-benchmarking tooling construction

FoodChain1028:
# Related Issue
#21 

# Detail
The details of this PR is illustrated in this [report](https://hackmd.io/yrKOugWQQ5eF6Ihr9EsOew?both).

# Brief Look of MSM running on CPU
msm benchmarks of time and memory usage

The raw figures are stored in [this chart](https://docs.google.com/spreadsheets/d/1cl96S-Cdh6YRIbwofVp_nyDngzsNXNhCy0X_yvhtcuw/edit?usp=sharing). We observed that the **allocated memory is higher** while **the number of msm running is lower**. And when the number went after approximately 200, allocated memory become stable.

(cc @moven0831)


moven0831:
### Key observations from the current benchmarks:

1. The average msm time stays constant ($\approx$ 5ms), and the total msm time grows linearly with the number of msm, suggesting Rust processes them sequentially.
2. The allocated memory usage become stable ($\approx$ 0.55 MiB) after performing 200 msm. We assume that there is some unused memory being freed by Rust and the remaining allocated memory is used by msm-related process.

Our thought is that how to speed up multiple msm operations (e.g. concurrency/multithreading) is not our priority. We will focus on how to accurately benchmark single MSM operation and improve its performance.

mopro-core/Cargo.toml
```diff
@@ -44,8 +44,14 @@ ark-zkey = { path = "../ark-zkey" }
 thiserror = "=1.0.39"
 color-eyre = "=0.6.2"
 criterion = "=0.3.6"
+jemalloc-ctl = "0.5.4"
```
oskarth, line 47: Can we hide these behind a feature flag? say `gpu-benchmarks`. This way we can be more aggressive with dependencies etc without bloating up library size for other consumers

mopro-core/src/middleware/mod.rs
```diff
@@ -1 +1,2 @@
 pub mod circom;
+pub mod gpu_exploration;
```
oskarth, line 2: Ditto re feature flag

mopro-core/src/middleware/gpu_exploration/msm_bench.csv
```diff
@@ -0,0 +1,35 @@
+num_msm,avg_processing_time(sec),total_processing_time(sec),memory_allocated(MiB)
```
oskarth, line 1: This probably belongs in a different non-code folder

maybe `mopro-core/benchmarks/gpu_explorations`?

mopro-core/src/middleware/gpu_exploration/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+#[derive(Debug, Clone)]
+pub struct BenchmarkResult {
+    pub num_msm: u32,
+    pub avg_processing_time: f64,
+    pub total_processing_time: f64,
+    pub allocated_memory: f64,
+}
+
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
```
oskarth, line 18: Would you mind adding a comment for why we add this?

oskarth (commented):
Thank you for the RP!

Left a few comments in-line. Some questions:

From https://github.com/oskarth/mopro/issues/21:

- Write basic [Swift test](https://github.com/oskarth/mopro/tree/main/mopro-ffi/tests/bindings) to check initial performance on laptop (not mobile/iOS)
- Add and call above function in the iOS example app [ViewController.swift](https://github.com/oskarth/mopro/blob/main/mopro-ios/MoproKit/Example/MoproKit/ViewController.swift)
- Run on a real iOS device to check performance
- Update README with instructions for how to run above; add initial benchmarks observed

Do you see these as being done in follow up PRs? My understanding is that this PR lays the groundwork for above by being able to see basic performance of many MSMs in Rust on a laptop.

mopro-core/Cargo.toml
moven0831: Done. I put them behind the `gpu-benchmarks` flag and revised related code in both `mopro-core` and `mopro-ffi`

mopro-core/src/middleware/gpu_exploration/msm_bench.csv
moven0831: Exactly, thanks for the suggestion

mopro-core/src/middleware/gpu_exploration/mod.rs
moven0831: After experiments, I discovered that this code will have the allocator record all the dynamic memory allocations. However, we only need to focus on processes related to MSM. I removed them to make the benchmarks more accurate.

moven0831: The new result shows that memory allocated for single MSM is $\approx{}0.05$ MiB

msm benchmarks of time and memory usage

moven0831:
Two major fix in the PR https://github.com/FoodChain1028/mopro-21/pull/2 :

1. all of the code in `gpu_explorations` is under the flag `gpu-benchmarks`
2. new benchmarks result is very different from the previous one (the main issue is that we did not use `--release` flag)
    - Average time for single MSM: $\approx{} 0.3$ ms
    - time for 10_000 MSM: $\approx{} 3,000$ ms
    - Average allocated memory for single MSM: $\approx{} 50,000$ Bytes
    
msm benchmarks of time and memory usage

moven0831:
> Thank you for the PR!
> 
> Left a few comments in-line. Some questions:
> 
> From #21:
> 
> * Write basic [Swift test](https://github.com/oskarth/mopro/tree/main/mopro-ffi/tests/bindings) to check initial performance on laptop (not mobile/iOS)
> * Add and call above function in the iOS example app [ViewController.swift](https://github.com/oskarth/mopro/blob/main/mopro-ios/MoproKit/Example/MoproKit/ViewController.swift)
> * Run on a real iOS device to check performance
> * Update README with instructions for how to run above; add initial benchmarks observed
> 
> Do you see these as being done in follow up PRs? My understanding is that this PR lays the groundwork for above by being able to see basic performance of many MSMs in Rust on a laptop.

Yes, both swift test and benchmarking on the real iOS device will be in our following PRs. The current PR is about the result of milestone 1 in https://github.com/privacy-scaling-explorations/acceleration-program/issues/29.

oskarth (approved):
Thanks for all the fixes, LGTM! Congratulations on your first PR :)

oskarth:
Just waiting for CI to pass, then we can merge

PR #54, opened 2024-01-15
fix(ios): fix package name

vivianjeng:
- fix uniffi-0.25 linker problem https://github.com/mozilla/uniffi-rs/issues/1922
- `uniffi-bindgen` should be overwritten if `uniffi` version updated
   https://github.com/oskarth/mopro/blob/2afe32041f89070e6996da8b38ec192fcd8dfe5f/scripts/prepare.sh#L136-L142


oskarth (approved):

oskarth:
I see, so we need uniffi-bindgen 0.25 https://github.com/mozilla/uniffi-rs/issues/1922#issuecomment-1893007056

PR #63, opened 2024-01-29
fix(generate_benchmark_report): main function exists even when feature is not enabled

moven0831:
FIx of the issue in https://github.com/oskarth/mopro/issues/61#issue-2104544124.

The main problem is that `./scripts/build_ios.sh simulator release` run `cargo build --release`. The build process can now be successfully executed and so is the script.


oskarth (approved):
Thanks

PR #67, opened 2024-01-31
feat: Use TOML for config

oskarth:
This PR starts to introduce a more robust way of configuring relevant build options and environment.

Currently there's a big surface area for options, and we still have some hardcoded leftovers. This means we currently have to do things like this: https://github.com/oskarth/mopro/pull/66/files/8e7d9dafaf0538ab84d739ec8a9f42437772df1a#diff-10095d80195138814b66f1d6f5b769b0763cd537d9ef563a28c6d75444a151d5 which makes things very confusing, especially when we have dealing with multiple circuits.

## Big idea

Specify options in a toml file:

```
# config-example.toml

[build]
device_type = "simulator" # Options: x86_64, simulator, device
build_mode = "release"    # Options: debug, release

[circuit]
dir = "examples/circom/keccak256" # Directory of the circuit
name = "keccak256_256_test"       # Name of the circuit

[dylib]
use_dylib = false        # Options: true, false
name = "keccak256.dylib" # Name of the dylib file, only used if use_dylib is true
```

Then call this with build scripts like so:

`./scripts/build_ios.sh config-example.toml`

It is then easy to introduce multiple configs (e.g. for Anon-Aadhaar), to debug etc.

TOML config is also a good way to prepare for Rust API. I made it work for shell scripts too since that's the current base.

## TODO before merge
- [ ] Sanity check of initial API (feedback welcome)
- [x] Get rid of hardcoded stuff in build.rs and use TOML config instead
- [x] Update Android script
- [ ] Make sure CI passes
- [x] Update README

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Prepare CI for iOS
         run: ./scripts/prepare_ci.sh
       - name: Build for iOS
-        run: ./scripts/build_ios.sh x86_64 debug
+        run: ./scripts/build_ios.sh config-example.toml
       - name: Run iOS tests
         run: |
           cd mopro-ios/MoproKit/Example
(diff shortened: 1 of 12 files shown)


oskarth:
Later on we can also add arrays to build for multiple platforms etc... starting simple though and just replacing existing positional argument and hardcoded stuff.

PR #80, opened 2024-02-24
chore: Add Cargo.toml workspace

oskarth:
Useful for including externally.


oskarth:
FYI @vivianjeng need to adjust `target` dir to top level for Android. I didn't do it in this PR because I didn't want to mess with that config right now 😅

PR #106, opened 2024-03-12
Feat/example dylib 4

oskarth:
Most dylib work has been merged already, see  recent commits, cli/build_ios script, and example app.

With this config and `mopro build --platforms=ios`:

1) dylib is included in a separate XCFramework
2) Circuit dylib is copied into app bundle as part of build process
3) Example works on real device detached from cable

Leaving as draft because I want to make sure integration works well for both dylib and non-dylib case. Might require user to manually add XCFramework in Xcode.

Screenshot 2024-03-12 at 16 10 38

I have not been able to test submitting to app store (Product -> Archive -> Validate/Upload) as I don't have a valid Apple Developer account, and won't be able to set it up properly quickly.

If someone can try it out and verify that it works that'd be much appreciated!

This PR replaces and improves on:
- https://github.com/oskarth/mopro/pull/7
- https://github.com/oskarth/mopro/pull/37
- https://github.com/oskarth/mopro/pull/38

(Note that there might be some minor regressions for non-dylib case from previous commits due to cache etc.)

### DIFF
--- a/mopro-ios/README.md
+++ b/mopro-ios/README.md
@@ -59,3 +59,10 @@ There are two ways to run tests for the example app:
    * `simulator`: running on the simulator (default)
    * `device`: running on a real IOS device
 4. Open `MoproKit/Example/MoproKit.xcworkspace` in Xcode and `cmd + R` for building
+
+## Dylib support
+
+When building for real devices on iOS we have to convert the `wasm` witness generation file to a `dylib`.
+When doing so, set the `use_dylib` flag to true in `mopro-config.toml`.
+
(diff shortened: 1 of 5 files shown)


oskarth:
Update: doesn't work for actual app store upload due to:

> ITMS-90426: Invalid Swift Support - The SwiftSupport folder is missing. Rebuild your app using the current public (GM) version of Xcode and resubmit it.

Which is related to dylib. After trying many different things, wee currently don't have a good way to approach this problem :/

https://forums.swift.org/t/swift-package-with-dynamic-libraries-leads-to-app-store-validation-error/66857/12

PR #111, opened 2024-03-13
fix(example-app): Adding Info.plist and icons to example app

0xturboblitz:
This allows for App Store and Test Flight deployment.

Also removing iPad from targets.

This solves https://github.com/oskarth/mopro/issues/108


oskarth:
Awesome, thank you! Later on we can replace icons with mopro logo, but for now proof of passport is good :D

PR #113, opened 2024-03-14
fix(core,scripts): Disable Circom optimizations

oskarth:
This seems to work better for some scenarios. Also delete old script.


vivianjeng:
we don't compile circuits in mopro anymore

PR #121, opened 2024-04-01
fix(cli,core): Fix first mopro deps error

oskarth:
https://github.com/oskarth/mopro/issues/119

- Default to simpler circuit
- Download artifacts to simulate mopro prepare

### DIFF
--- a/mopro-core/build.rs
+++ b/mopro-core/build.rs
@@ -71,12 +71,12 @@ fn read_config() -> Result<Config> {
             println!("cargo:warning=BUILD_CONFIG_PATH not set. Using default configuration.");
             let default_config = r#"
                 [circuit]
-                dir = "examples/circom/keccak256"
-                name = "keccak256_256_test"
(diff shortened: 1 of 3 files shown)


oskarth:
@vivianjeng do you know why it doesn't work on Android? Assumption re keccak256 default config or something?

mopro-core/build.rs
```diff
@@ -71,12 +71,12 @@ fn read_config() -> Result<Config> {
...
                 [circuit]
-                dir = "examples/circom/keccak256"
-                name = "keccak256_256_test"
+                dir = "examples/circom/multiplier2"
+                name = "multiplier2"
 
                 [dylib]
                 use_dylib = false
-                name = "keccak256.dylib"
+                name = "multiplier2.dylib"
```
vivianjeng, line 79: it looks like the mopro-core tests takes `multiplier2` for `test_generate_proof2`
so the inputs and outputs don't match keccak256 proof

PR #122, opened 2024-04-01
Reproducing Missing Bundle Executable error using framework instead of xcframework

0xturboblitz:
I’m not sure it’s much progress, but the error is different now:
```
ITMS-90868: Missing Bundle Executable - You must include a valid CFBundleExecutable key in your bundle's information property list file for bundle 'ExampleApp.app/Frameworks/CircuitBindings.framework'.
```
Instead of:
```
ITMS-90426: Invalid Swift Support - The SwiftSupport folder is missing. Rebuild your app using the current public (GM) version of Xcode and resubmit it.
```


vivianjeng:
dylib feature is removed

PR #123, opened 2024-04-02
fix: fix dylib file path in build scripts

vivianjeng:
- fix copying `.dylib` from `mopro-core/target`
- fix duplicated `.dylib` in script and `config.toml`


oskarth (approved):
Hm, weird, I feel like I've done similar changes before and got it to work from a clean slate. But maybe I missed something. If this resolves some problem for you OK let's merge.

PR #125, opened 2024-04-05
fix: fix npm_install script if there is no package.json

vivianjeng:
- for multiplier.circom, there is no `package.json` so it throws error.
- In `prepare.sh` we can check if there is a `package.json` file then execute `npm install`


oskarth (approved):
Uhm right, good catch

Issue #25, opened 2023-10-31
Improve zkey load time


oskarth:
# Problem

It is slow to load zkey.

`cargo test --release test_generate_proof2 -- --nocapture` shows it takes ~70-80s to load in zkey for a ~80mb file (Keccak256).

Zkey is ~80mb but would expect this to be much faster.

This seems to be the case regardless of if we are using dylib or not.

## Approaches

Two ideas:

1) Figure out how to improve in terms of reading in the zkey file (maybe memmap or something, but this adds complexity for iOS integration)
2) Do this separately and persist arkworks ProvingKey in a better way (e.g. serialize to disk beforehand)

# Acceptance criteria

Much faster loading of zkey, or other way to improve this type of load up time

oskarth:
I confirm that this is a problem on iOS device too. My hunch is this will be improved using dylib.

Still seems weird to me this is so slow though.

oskarth:
Same basic behavior. With this we isolate this to initializing of library.

https://github.com/oskarth/mopro/blob/main/mopro-core/src/middleware/circom/mod.rs#L87 for timing

Gonna leave this for now but two approaches:

1) Figure out how to improve in terms of reading in the zkey file
2) Do this separately and persist arkworks ProvingKey in a better way (e.g. serialize to disk beforehand)

oskarth:
We should be able to preprocess zkey and then quickly load proving key and matrices from disk:

```
// Reads a SnarkJS ZKey file into an Arkworks ProvingKey.
pub fn read_zkey<R: Read + Seek>(
    reader: &mut R,
) -> IoResult<(ProvingKey<Bn254>, ConstraintMatrices<Fr>)> {
    let mut binfile = BinFile::new(reader)?;
    let proving_key = binfile.proving_key()?;
    let matrices = binfile.matrices()?;
    Ok((proving_key, matrices))
}
```

oskarth:
I tried out some things here https://github.com/oskarth/mopro/pull/26 and it seems like the slow part is the ProvingKey. I think this is because it is doing a bunch of EC operations etc, so there might be a more efficient way to do serialization field by field.

```
running 1 test
Reading zkey from: ../mopro-core/examples/circom/keccak256/target/keccak256_256_test_final.zkey

test tests::test_serialization_deserialization has been running for over 60 seconds
Time to read zkey: 99.439572834s
[...]

PR #129, opened 2024-04-21
feat(core): update read_zkey with `new_unchecked`

vivianjeng:
According to https://github.com/zkmopro/mopro/issues/25#issuecomment-2067880755

| loading zkey time | ark-zkey | with `new_unchecked` |
|:--:|:--:|:--:|
| multiplier2 | 872.00ns |  1.21µs |
| keccak256 | 115.82s |  1.56s |


oskarth:
I wonder if we want to combine ark-zkey with this? Because with ark-zkey we get 50% file size reduction due to point compression.

mopro-core/src/middleware/circom/mod.rs
```diff
@@ -66,21 +67,21 @@ impl Default for CircomState {
...
 
-const ARKZKEY_BYTES: &[u8] = include_bytes!(env!("BUILD_RS_ARKZKEY_FILE"));
+// const ARKZKEY_BYTES: &[u8] = include_bytes!(env!("BUILD_RS_ARKZKEY_FILE"));
 
-// static ZKEY: Lazy<(ProvingKey<Bn254>, ConstraintMatrices<Fr>)> = Lazy::new(|| {
-//     let mut reader = Cursor::new(ZKEY_BYTES);
-//     read_zkey(&mut reader).expect("Failed to read zkey")
-// });
-
-static ARKZKEY: Lazy<(ProvingKey<Bn254>, ConstraintMatrices<Fr>)> = Lazy::new(|| {
```
oskarth, line 78: do we have to remove the ARKZKEY stuff you think? or is this preferable?

thinking mostly about space savings which are quite significant

oskarth (approved):

oskarth:
If we remove/don't update arkzkey stuff, could we create an issue explaining problem and current state? So even if we don't do it now we can introduce this for arkzkey too

vivianjeng:
Create an issue here
https://github.com/zkmopro/mopro/issues/130

PR #138, opened 2024-05-18
(fix: cli) Fixed issue with IOS debug mode not able to build

ElusAegis:
Fixed a type in `mopro/scripts/cli/build_ios.sh` that prevents running `mopro build --platforms ios` when configured for build mode.


vivianjeng (approved):
Thank you!

PR #152, opened 2024-06-04
feat: enable parallel in gpu-benchmarks

vivianjeng:
- enable parallel feature https://github.com/Trapdoor-Tech/TrapdoorTech-zprize-mobile-harness/blob/master/Cargo.toml#L24
- result (on simulator)

   (on iPhone 12 mini)
    IMG_6571


moven0831 (approved):
LGTM!

PR #156, opened 2024-06-11
Integrate metal msm in ios

FoodChain1028:
Integrating metal msm into `research/gpu-exploration-app`.

Things that also have to improve in this PR
* [X] configuration when compiling `metallib` (in the file we made 2 targets `ios` and `mac`, which respectively represents target `iphoneos` and `macosx`)
* [ ] fix the GPU hang error running in mobile device (current test targets with iphone 14 pro (6GB memory) and ipad pro 3rd generation (8GB memory) both occur **GPU hang error**).

Currently we encounter a problem that should be solved. When the instance size become larger, `GPU hang error` happens.


doutv:
Runtime error on iPad 15.2. 

Seems that it requires iOS 17+. 
https://developer.apple.com/documentation/metal/mtllanguageversion/version3_1
```
Running MSM in algorithm: Metal Msm (our work)...
Vectors already generated
Init metal (GPU) state...
thread '<unnamed>' panicked at mopro-core/src/middleware/gpu_explorations/metal/msm.rs:65:67:
called `Result::unwrap()` on an `Err` value: PipelineError("Function initialize_buckets is using language version 3.1 which is incompatible with this OS.")
```

moven0831:
This PR will be moved with a newer PR for only iOS integration.

PR #158, opened 2024-06-11
Rust formatting fix for `templates/mopro-example-app/core`

ElusAegis:
Before the `fmt` ci was not checking the issues in the `templates/mopro-example-app`, which resulted in the code not being correctly formatted error when running `cargo fmt  --all -- --check`.

**This PR:**
1. Fixed the code formatting issue.
2. Added the additional check to the ci to catch similar issues.


ElusAegis:
This PR's cli CI fails, even though the PR does no breaking changes, rather reordering the code. Do not understand why. Managed to run the same IOS test where the issue occures successfully on my local machine.

Issue #151, opened 2024-06-03
Support latest rust toolchain


vivianjeng:
# Problem 
Now the latest rust toolchain version cannot run circom-compat successfully
https://github.com/zkmopro/mopro/issues/148#issuecomment-2144588600

## Details
we should either
1. change circom-compat to fit the latest toolchain
    - reference for update circom-compat: https://github.com/l-adic/circom-compat
2. remove wasmer dependencies since it cannot be run on iOS and cannot be submitted to testfligt 
    - reference #107 #127 

## Acceptance criteria
Remove the specific toolchain version in [build-and-test.yml](https://github.com/zkmopro/mopro/blob/main/.github/workflows/build-and-test.yml) and CI should succeed

Issue #107, opened 2024-03-13
Dylib Framework not accepted by App Store


oskarth:
# Problem

When submitting example app to app store we get the following error:

```
We noticed one or more issues with a recent delivery for the following app:

MoproExampleApp
Version 1.0
Build 1
Please correct the following issues and upload a new binary to App Store Connect.

ITMS-90426: Invalid Swift Support - The SwiftSupport folder is missing. Rebuild your app using the current public (GM) version of Xcode and resubmit it.

Apple Developer Relations
```

This only happens when you do app submission, not when building the app locally.

## Details

Brain dump of relevant things.

See https://github.com/oskarth/mopro/pull/106 for latest PR. This is the circuit dylib in an XCFramework and then embedded in app. https://github.com/oskarth/mopro/blob/main/scripts/cli/build_ios.sh#L198-L217

Dylib support https://github.com/oskarth/mopro/issues/11 issue.

We load circuit by path, so we need dylib path. If we naively convert to framework binary format with lipo it doesn't seem to be able to read. Problem doesn't mention dylib, but when removing it it works.

Some relevant links:

https://developer.apple.com/documentation/bundleresources/placing_content_in_a_bundle

> Avoid using dynamic library files (.dylib files) for dynamic linking. An XCFramework can include dynamic library files, but only macOS supports these libraries for dynamic linking. Dynamic linking on iOS, watchOS, and tvOS requires the XCFramework to contain .framework bundles.

https://forums.swift.org/t/swift-package-with-dynamic-libraries-leads-to-app-store-validation-error/66857/12

## Acceptance criteria

Successful submission to TestFlight/App store with circuit as dylib embedded.

Alt: Some other way of doing witness generation (see other issues)

oskarth:
Notes:

![圖片](https://github.com/oskarth/mopro/assets/1552237/4c199d5f-1f6f-4ffa-bc76-6926cff9608e)

---

To try (from Phil):

https://medium.com/trueengineering/xcode-and-xcframeworks-new-format-of-packing-frameworks-ca15db2381d3

![圖片](https://github.com/oskarth/mopro/assets/1552237/e2c30d5a-16b6-4ddc-aad6-2896c8e4ea8a)

0xturboblitz:
Here is where I am right now. Currently I'm working on my repo, which has a bit of a different structure, but not that different.
[...]

PR #159, opened 2024-06-12
wasmer 4.x support

vimwitch:
This PR bumps the `wasmer` version from 2.x to 4.x. This relies on updates to the `circom-compat` repo to support this version (implemented in the fork [here](https://github.com/zkmopro/circom-compat)).

Related #151


vivianjeng:
There is a good implementation example here
https://github.com/l-adic/circom-compat
but for iOS we still need to patch the wasmer like
https://github.com/wasmerio/wasmer/commit/09c7070

I haven't tried either if we patched the wasmer it works for iOS
or if it can be published in testflight https://github.com/zkmopro/mopro/issues/107

vivianjeng:
You can also fork the latest wasmer in zkmopro and patch these lines

vimwitch:
@vivianjeng I think this is ready for review now. I removed the singleton pattern that was caching the wasm executor because of changes in wasmer@4. There's an in memory `Store` that tracks execution of the wasm and needs to be mutable so it can't be in a static singleton.

The new wasmer also relies more on async functions so tests are async.

vivianjeng:
I tried to run it on iOS
Now it can be run if the device is connected to xcode
but it cannot be run if the device is disconnected to xcode
(you download the app, stop xcode, and open the app again and run)

Previously we use `dylib` feature to fix this
but I tried to use `dylib`, it throws
```sh
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: input file: /Users/zhengyawen/Documents/GitHub/vimwitch/mopro/mopro-example-app/target/aarch64-apple-ios/release/multiplier2.dylib is not a Mach-O file
```

Not sure if the wasmer is wrong
but in wasmer v4, there is no dylib feature

I saw there is also a solution like https://github.com/oliveeyay/SwiftWasmer
https://github.com/wasmerio/wasmer/issues/4343
I am not sure if it works in our case

Do you know if the `dylib` feature can be fixed?
Is it possible that it can be run when disconnected to xcode?

vimwitch:
Closing in favor of #179

PR #168, opened 2024-06-20
(feat)core: Update to Halo2 circuit Interface

ElusAegis:
This PR focused on improving the interface which `mopro-core` uses to use user-provided halo2 circuits.

1. **Changes in `halo2/mod.rs`**: Because the circuits can be built on top of `halo2_proofs` which uses the `params` feature, while some do not, the decision was to move the variable logic to the user-provided circuit. 
2. **Introduced `mopro.rs` in `fibonacci`**: To simplify the `lib.rs` as well as make it easier for people integrating the Halo2 version of **mopro**, moved all functionality which `mopro-core` requires for users circuit crate to implement into a dedicated `mopro.rs` module file.
3. **`mopro-core/build.rs`**: Added Halo2 specific template config to make testing possible for halo2 template circuits within `mopro-core`
4. **template** changes: 
    - Updated the Halo2 Fibonacci example in template to be consistent with the new interface.  
    - Renamed Halo2 Finabocci View to be explicitly for Fibonacci
5. Cosmetic changes to documentation / descriptions.


vivianjeng (approved):
LGTM 
thank you

PR #169, opened 2024-06-20
(fix)build: Enable changing Halo2 Circuits

ElusAegis:
PR to fix an issue of being unable to switch between two different Halo2 circuits.

1. **`mopro-core/build.rs`**: remove the old code to create `.cargo/config.toml` from the build script because the changes to `.cargo/config.toml` must be made before the `cargo build` is invoked.
2. **`scripts/.../build_ios.sh`**: introduced the code to create/update `.cargo/config.toml` to be early enough

### Future Work:
Update other platform build scripts.


mopro-core/build.rs
```diff
@@ -370,7 +292,7 @@ fn main() -> color_eyre::eyre::Result<()> {
 
         build_halo2_circuit(&circuit_dir_path, circuit_name)?;
 
-        update_halo2_cargo_config(&circuit_dir_path);
+        // update_halo2_cargo_config(&circuit_dir_path);
```
vivianjeng, line 295: this can be removed?

PR #170, opened 2024-06-20
(feat)examples: Halo2 Keccak256 Example Circuit

ElusAegis:
1. Added a Keccak256 Circuit to `mopro-core/examples/halo2/keccak256` and `templates/core/circuits/halo2-keccak256`.
2. Added an option to run a `halo2-keccak` circuit to the template config.
3. Added a Swift view for `halo2-keccak` circuit.


vivianjeng (commented):
when I execute `cargo test -- --nocapture` in `mopro-core/examples/halo2/keccak256`
it shows
```sh
(base) ➜  mopro-halo2 git:(feat/halo2-keccak) cd mopro-core/examples/halo2/keccak256 
(base) ➜  keccak256 git:(feat/halo2-keccak) cargo test -- --nocapture
error: current package believes it's in a workspace when it's not:
current:   /Users/zhengyawen/Documents/GitHub/mopro-halo2/mopro-core/examples/halo2/keccak256/Cargo.toml
workspace: /Users/zhengyawen/Documents/GitHub/mopro-halo2/Cargo.toml

this may be fixable by adding `mopro-core/examples/halo2/keccak256` to the `workspace.members` array of the manifest located at: /Users/zhengyawen/Documents/GitHub/mopro-halo2/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```
maybe the instruction in the readme is not updated?
or the workspace settings are wrong?

and when I used mopro-cli to run `mopro prepare` with keccak256 circuits
it shows the similar error messages

and I am not familiar with halo2
does it require to copy all the files (e.g. `util/`, `vanilla/`...) in the `examples/halo2`?
or we can use a repo to host all the common circuits (e.g. `fibonacci`, `keccak256`...)
and use the functions in `examples/halo2` like
```toml
[package]
name = "halo2-keccak256-mopro"
version = "0.1.0"
edition = "2021"

[dependencies]
halo2-keccak256 = { git =  "https://github.com/..."}
```
and just keep the file `src/mopro.rs`

ElusAegis:
The situation with `workspace` is quite tough.

Both `keccak` and `fibonacci` crates are actually called `halo2-circuit`, as the idea is that whenever anyone wants to add another circuit to the mix, they just create a crate called `halo2-circuit`, provide a path to it in the config, and it will overwrite it in the dependencies (a hack way to include a dynamic build dependency, I looked for a better way, but could not find anything).

And the idea behind the `examples/halo2` was to provide an example of such substitution. However, apparently, as you can not have two same-named workspace members, both examples need to be either their own workspace or one of them should be in `mopro` workspace, one on its own.

At the same time, if we include a local dependency in another create, that is located inside of the another crate's workspace, Cargo enforces it to be in a workspace, and if it is its own workspace, it throws an error. However, if it is a path override, it is okay...

So the following should work:
- `fibonacci` example to be added to the main workspace
- `keccak` example to be as its own workspace

vivianjeng:
@ElusAegis how about also setting them different feature flags
e.g. `fibonacci` and `keccak256`
so we enable different circuits by enabling different feature flags

ElusAegis:
I also thought about this. One option is to do place both samples inside of one crate crate, and use features to select between them, similar to FFI. However, that would make the examples complicated and also not quite right, as instead of two independent crates we will get just one shared. This is not ideal as the whole point of examples what in my opinion to show how to have separate crates and do paths overrides.

Another option is to overwrite the dependency in the 'mopro-core' based on the feature flag. However, that seems impossible if both crates have the same name. 

I propose to create an issue for that, and figure out something later - if there is a better solution.

vivianjeng:
@ElusAegis one other idea from @vimwitch is that
we just use the `srs`, `pk`, `vk` in each example
you can maintain the halo2 circuits in either your repo or zkmopro repo
and write down in readme or doc: https://zkmopro.org/docs/halo2/introduction
how to generate the `srs`, `pk`, `vk`
so that we don't have to maintain the halo2 circuits in mopro

ElusAegis:
I am not sure I understand. The issue is not just in the circuit assets, like `srs`, `pk, `vk`, but also the logic to generate the proof. It is custom for each circuit afaik (may be there is a way not to have it like that, which I am not aware). So when a user calls `generateStaticProof`, it should pick up a custom logic implemented by each circuit crate

mopro-core/examples/halo2/keccak256/src/mopro.rs
```diff
@@ -0,0 +1,113 @@
+use crate::circuit::{pack_input_to_instance, unpack_input, KeccakCircuit};
+use crate::vanilla::KeccakConfigParams;
+use crate::{CIRCUIT_PARAMS, K, ROWS_PER_ROUND};
```
vivianjeng, line 3: I see what you mean of the logic to generate proof.

but can we not using `crate::circuit` but instead using like `halo2_keccak256_circuit::circuit...`
and you can maintain `halo2_keccak256_cirucit` in other repo
so we don't have to maintain the copy of keccak circuits

ElusAegis: Ah, I think I get it - the idea is to make the Keccak circuit leaner, right? We absolutely can do this, such as move the bulk of the logic somewhere else, and only use `halo2-circuit` to have an adapter from the `keccak` to work with `mopro` - effectively just keep the `mopro.rs` module as the rest can be placed outside. 

Do not think this will resolve the issue we have with the workspaces, but it will indeed simplify the code. Is this what you are suggesting this for?

vivianjeng: yes
so maybe we can easily control halo2 circuits by halo2 features
e.g. `--features halo2-fibonacci`, `--features halo2-keccak`

PR #173, opened 2024-06-21
(feat)ffi: Refactored FFI 

ElusAegis:
The following was done:
1. Added a `circom` feature 
2. Defined a common set of functions that each adapter should implement. 
3. Improved how each adapter is exported by using `lib.rs` and choosing each adapter at compile time based on features.
4.  Renamed Halo2 `proof` and `verify` functions to be commonly named with the `circom` static version.
5.  Added a test for `halo2` ffi side, and added it to the ci as well.
6. Updated resulting template bindings as well as Halo2 sample swift code.


.github/workflows/build-and-test.yml
```diff
@@ -123,12 +123,18 @@ jobs:
...
-            - name: Run FFI tests
+            - name: Run FFI circom tests
               # TODO: Fix this custom jar thing
               run: |
                   cd mopro-ffi/
                   curl -L https://repo1.maven.org/maven2/net/java/dev/jna/jna/5.13.0/jna-5.13.0.jar -o jna-5.13.0.jar
                   CLASSPATH=jna-5.13.0.jar cargo test -- --nocapture
+            - name: Run FFI halo2 tests
+              run: |
+                  export BUILD_CONFIG_PATH="$(pwd)/config-halo2.toml"
```
vivianjeng, line 134: there is no `config-halo2.toml` after merging the pervious PRs
please check the test 
thank you

vivianjeng (changes requested):

.github/workflows/build-and-test.yml
ElusAegis: Thank you, fixed

PR #179, opened 2024-06-22
refactor: remove wasm, add native proof generation

vimwitch:
This PR removes the wasm logic as well as the `dylib` and `calc-native-witness` logic. Instead it relies on the [`rust-witness`](https://github.com/zkmopro/rust-witness) package to natively build witnesses. This happens by taking the wasm, compiling to c, then using a macro to automatically write a function that calls the c function to build the witness.

The UDL bindings are very complex right now. Instead of passing paths between the rust side and the mobile side I think it makes more sense to build a dictionary on the rust side for each circuit. Imagine there is a build with 2 circuits: `multiplier2` and `keccak256`. Instead of having `initialize(zkey_path, wasm_path)` we could have `initialize(circuit_name)`, and define the paths to zkeys, witnesses, etc at compile time in rust - which should already be happening. Then we have minimal complexity moving over the bridge.

Instead of giving the user a config file with strings we could give them a config file as a rust source file. This way they can define their own logic for calculating witnesses, loading zkeys, etc. In practice they would probably use whatever loaders we recommend.

[This](https://github.com/zkmopro/mopro/blob/8d08737e78b239891c9154aa6964953af1031969/mopro-ffi/src/circom.rs#L119) is what such a config function might look like.

### DIFF
--- a/ark-zkey/Cargo.toml
+++ b/ark-zkey/Cargo.toml
@@ -23,7 +23,7 @@ flamer = "0.5"
 ark-serialize = { version = "=0.4.1", features = ["derive"] }
 ark-bn254 = { version = "=0.4.0" }
 ark-groth16 = { version = "=0.4.0" }
-ark-circom = { git = "https://github.com/vimwitch/circom-compat.git", branch = "wasm-delete", version = "0.1.0" }
+ark-circom = { git = "https://github.com/zkmopro/circom-compat.git", branch = "wasm-delete", version = "0.1.0" }
 ark-relations = { version = "=0.4.0" }
 ark-ff = { version = "=0.4.1" }
 ark-ec = { version = "=0.4.1" }
--- a/ark-zkey/Cargo.toml
+++ b/ark-zkey/Cargo.toml
@@ -23,7 +23,7 @@ flamer = "0.5"
 ark-serialize = { version = "=0.4.1", features = ["derive"] }
 ark-bn254 = { version = "=0.4.0" }
 ark-groth16 = { version = "=0.4.0" }
-ark-circom = { git = "https://github.com/vimwitch/circom-compat.git", version = "0.1.0" }
+ark-circom = { git = "https://github.com/vimwitch/circom-compat.git", branch = "wasm-delete", version = "0.1.0" }
 ark-relations = { version = "=0.4.0" }
 ark-ff = { version = "=0.4.1" }
 ark-ec = { version = "=0.4.1" }
--- a/mopro-core/build.rs
+++ b/mopro-core/build.rs
@@ -183,6 +183,7 @@ fn main() -> color_eyre::eyre::Result<()> {
         // If Circom feature is enabled, build Circom Circuit
         println!("cargo:warning=Building Circom circuit...");
 
+        // see here: https://github.com/zkmopro/rust-witness/?tab=readme-ov-file#rust-witness
         transpile_wasm(String::from("./examples"));
     }
 
--- a/mopro-core/build.rs
+++ b/mopro-core/build.rs
@@ -3,15 +3,9 @@ use color_eyre::eyre::Result;
 
 use serde::Deserialize;
 
-use enumset::enum_set;
-use enumset::EnumSet;
 use rust_witness::transpile::transpile_wasm;
-use serde_derive::Deserialize;
-use std::fs::File;
-use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::{env, fs};
-use toml::Value;
 
 #[derive(Deserialize)]
 struct Config {
--- a/mopro-core/build.rs
+++ b/mopro-core/build.rs
@@ -2,15 +2,20 @@ use color_eyre::eyre::eyre;
 use color_eyre::eyre::Result;
 
 use serde::Deserialize;
+
+use enumset::enum_set;
+use enumset::EnumSet;
+use rust_witness::transpile::transpile_wasm;
+use serde_derive::Deserialize;
+use std::fs::File;
+use std::io::Write;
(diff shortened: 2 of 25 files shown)


ark-zkey/Cargo.toml
```diff
@@ -23,7 +23,7 @@ flamer = "0.5"
 ark-serialize = { version = "=0.4.1", features = ["derive"] }
 ark-bn254 = { version = "=0.4.0" }
 ark-groth16 = { version = "=0.4.0" }
-ark-circom = { git = "https://github.com/vimwitch/circom-compat.git", version = "0.1.0" }
+ark-circom = { git = "https://github.com/vimwitch/circom-compat.git", branch = "wasm-delete", version = "0.1.0" }
```
vivianjeng, line 26: I think we can use `zkmopro/circom-compat`

mopro-core/build.rs
```diff
@@ -375,12 +245,7 @@ fn main() -> color_eyre::eyre::Result<()> {
         // Otherwise, we default to Circom circuit
         println!("cargo:warning=Building Circom circuit...");
 
-        build_circom_circuit(&circuit_dir_path, circuit_name)?;
-
-        #[cfg(feature = "build-native-witness")]
-        build_witness_graph()?;
-
-        build_dylib(&config)?;
+        transpile_wasm(String::from("./examples"));
```
vivianjeng, line 248: what does the `./examples` here mean?

mopro-core/src/middleware/circom/serialization.rs
```diff
@@ -66,24 +70,30 @@ mod tests {
...
         r1cs_path: &str,
     ) -> Result<SerializableProvingKey, MoproError> {
-        assert_paths_exists(wasm_path, r1cs_path)?;
+        assert_paths_exists(r1cs_path)?;
 
-        let cfg = CircomConfig::<Bn254>::new(wasm_path, r1cs_path)
-            .map_err(|e| MoproError::CircomError(e.to_string()))?;
+        // TODO: cleanup the circom-compat implementation
+        // TODO: clean up error handling
+        // why do we have a MoproError type?
```
vivianjeng, line 87: @oskarth

scripts/prepare_ci.sh
```diff
@@ -111,7 +111,11 @@ compile_circuit keccak256 keccak256_256_test.circom
 
 # Setup and compile rsa
 npm_install rsa
-compile_circuit rsa main.circom
+compile_circuit rsa main_rsa.circom
+
+# # Setup and compile anonAadhaar
+npm_install anonAadhaar
+compile_circuit anonAadhaar aadhaar-verifier.circom
```
vivianjeng, line 118: we skip anon aadhaar tests in CI because it takes too much time

vivianjeng (commented):
I tried to execute `mopro build --platforms ios`
but it took so long (like CI)
```sh
Compiling ark-circom v0.1.0 (https://github.com/vimwitch/circom-compat.git?branch=wasm-delete#2651daab)
   Compiling ark-zkey v0.1.0 (/Users/zhengyawen/Documents/GitHub/vimwitch/mopro/ark-zkey)
^Z  Building [=======================> ] 402/407: mopro-ffi(build), mopro-core(build) 
```
do you know why?

and we now have feature flag for circom
https://github.com/zkmopro/mopro/blob/9ded18c0e36b5aa959e5876c90502d5d57f1788b/mopro-core/Cargo.toml#L19
please also help to update this
thank you

mopro-core/build.rs
vimwitch: It's referencing the directory `mopro-core/examples`. It's telling `rust-witness` to look in that directory (recursively) for wasm files to transpile and link to the rust build. Any wasm files it transpiles will be available in rust for execution as C code.

mopro-core/src/middleware/circom/serialization.rs
vimwitch: Ah that comment is old, I believe the `MoproError` exists for compatibility with Uniffi but could use more clarification.

scripts/prepare_ci.sh
vimwitch: Right, I uncommented to build it and test witness generation.

vimwitch:
@vivianjeng I debugged the slow build time and it's because of the anon-aadhaar circuit. The C source code for the witness generator is 70 MB which takes quite a long time to compile. I put the aadhaar test behind a feature flag so we don't need need to build it.

vimwitch:
@vivianjeng The CLI tests are failing right now. To fix this we need to (at a minimum) add logic to bundle zkeys for each mobile platform. If we do that we can prove the pre-compiled circuits. To prove custom circuits we need to add the ability to configure via rust.

I think these items are both out of scope of this PR.

mopro-core/build.rs
vivianjeng: I think it is a little bit confusing
maybe you can add some comment messages above

mopro-core/src/middleware/circom/serialization.rs
vivianjeng: can you help delete the messages here (if it is not used)

vimwitch:
> I am also thinking about can we move the transpile_wasm to mopro prepare instead of mopro build? If we have large circuits it might take too much time to build/update bindings.

This is possible, but will expose a decent amount of complexity from rust-witness. We should instead look at caching more aggressively in the rust-witness package - possibly using a global cache with a static library keyed to the hash of the wasm file or something. But we can implement that without changing mopro.

vivianjeng (approved):
Perfect!

Issue #182, opened 2024-06-25
Roadmap to release


vimwitch:
- [ ] move mopro-core logic into mopro-ffi
  - [ ] delete mopro-core
  - [ ] delete [static/const embedded srs/zkey/vkey/etc.](https://github.com/zkmopro/mopro/blob/1560a44bbc2dd503953b71385b146b5bf7785484/mopro-core/src/middleware/halo2/mod.rs#L28)
    - [ ] accept path to constants when building proofs
- [ ] separate UDL interfaces for circom and halo2
  - [ ] remove conditional compilation for individual functions (#180)
- [ ] clean up examples
  - [ ] only store srs/zkey/vkey/wasm files - move the halo2/circom source to a different repo
- [ ] refactor cli
  - [ ] translate build_ios, build_android, build_core into rust (don't simply invoke shell scripts from rust)
    - [ ] delete TOML/mopro-config parsing logic
  - [ ] delete shell script cli implementation
  - [ ] delete circom build logic (delegate to mopro user)
  - [ ] export these functions as functions designed to be used in a build script
- [ ] write an example mobile app project in a separate repo
  - [ ] add a `build.rs` that calls the above rust implementations (build_ios, build_android, etc)
  - [ ] build mobile project using `cargo build` instead of `mopro build`
  - [ ] build lib_mopro_ffi.a from source, specify config in rust source
  - [ ] delete toml project config
  - [ ] copy zkeys/vkeys/srs/etc to app bundle
  - [ ] only use mopro cli for initializing projects
- [ ] build template from example project

vivianjeng:
>  delete mopro-core

We can just call it `mopro`?
if we will seperate `mopro-cli`
or just `src` if it is not monorepo anymore

I agree with the idea to combine mopro-core and mopro-ffi
they are too similar and not necessary to be seperated

>  delete TOML/mopro-config parsing logic

it will be controlled by feature flag or how to set the config?

>  delete shell script cli implementation

it means to implement with rust?

>  only use mopro cli for initializing projects

one other important features of mopro cli is to export bindings
if developers already built their apps (e.g. anon aadhaar)
instead of creating a new project, they might need only the `moproFFI` library
https://zkmopro.org/docs/circom/ios#getting-started-with-exported-bindings
if we can export `moproFFI` library without creating a new project it will be good
(or even we publish `moproFFI` to https://cocoapods.org/ it will be good too)

vivianjeng:
about 
> build template from example project

[...]

PR #183, opened 2024-06-25
tracking #182 (bob de boumaa rework)

### LINKED ISSUE #180: Namespace circom and halo2 seperately
We have a bunch of functions for generating proofs, serializing, etc. These functions are exposed in a global way to the mobile side and we're using compiler flags to swap which function is enabled. This is a bit of a mess when trying to parse the logic. It also prevents compiling with support for both circom and halo2 at the same time.

Instead we should namespace the halo2 and circom logic separately and use a single compiler flag to enable or disable the modules. Better yet, we could always compile support for both modules and avoid conditional compile logic altogether.

vimwitch:
- [x] move mopro-core logic into mopro-ffi
  - [x] delete mopro-core
  - [x] delete [static/const embedded srs/zkey/vkey/etc.](https://github.com/zkmopro/mopro/blob/1560a44bbc2dd503953b71385b146b5bf7785484/mopro-core/src/middleware/halo2/mod.rs#L28)
    - [x] accept path to constants when building proofs
- [x] separate UDL interfaces for circom and halo2
  - [x] remove conditional compilation for individual functions (#180)
- [x] clean up examples
  - [x] only store srs/zkey/vkey/wasm files
  - [ ] move the halo2/circom source to a different repo
- [ ] refactor cli
  - [ ] translate build_ios, build_android, build_core into rust (don't simply invoke shell scripts from rust)
    - [x] ios
    - [ ] android
  - [x] delete TOML/mopro-config parsing logic
  - [x] delete shell script cli implementation
  - [x] delete circom build logic (delegate to mopro user)
  - [x] export these functions as functions designed to be used in a build script
- [ ] write an example mobile app project in a separate repo
  - [x] add a `build.rs` that calls the above rust implementations (build_ios, build_android, etc)
  - [x] build mobile project using `cargo build` instead of `mopro build`
  - [x] build lib_mopro_ffi.a from source, specify config in rust source
  - [x] delete toml project config
  - [x] copy zkeys/vkeys/srs/etc to app bundle
  - [ ] only use mopro cli for initializing projects
- [ ] build template from example project

This PR transfers the proving logic from mopro-core to mopro-ffi. The `examples` directory is removed and the artifacts are committed directly into `mopro-ffi/test-vectors`. All the circuits except for `multiplier2` and `keccak256` have been removed.

Proving functions are now:
- `generate_circom_proof`
- `verify_circom_proof`
- `generate_halo2_proof`
- `verify_halo2_proof`

The halo2/circom logic is also separated so they can be enabled at the same time.

Related #182 
Closes #180


oskarth (changes requested):
Thanks for spike to integrate rust-witness.

However, this PR is too big with a massive surface area, impossible to review. It is also not actually targeting any specific identified issue, and probably contains 5-10 different issues, that may or may not be a problem.

See message in group chat.

Issue #153, opened 2024-06-07
Optimize Metal MSM Implementation for Enhanced GPU Utilization


moven0831:
# Problem
The current implementation of Metal MSM leverages GPU resources only partially, leading to suboptimal performance. To fully exploit GPU capabilities, we need to optimize the MSM process by addressing identified overheads and implementing enhancements.

# Details
### Current Implementation
The Metal MSM implementation currently derives from [Lambdawork's metal backend](https://github.com/lambdaclass/lambdaworks/tree/metal-pippenger/gpu/src/metal/shaders) and [Arkworks' MSM implementation](https://github.com/arkworks-rs/algebra/blob/master/ec/src/scalar_mul/variable_base/mod.rs#L157-L252). Metal is used only for the MSM accumulation phase on the GPU, while other phases run on the CPU. This partial GPU utilization limits potential performance gains.

### Identified Overheads
1. Parsing points and scalars for GPU computation.
2. Converting GPU-calculated buckets back to Arkworks-compatible types.

### Optimization Goals
1. **Full GPU Implementation**: Implement the entire MSM process on the GPU to reduce data conversion overhead by only parsing the final MSM result back to Arkworks-compatible types.
2. **Initialization Speed**: Accelerate the initialization of points and scalars for GPU computation.
3. **Advanced Techniques**: Apply optimization techniques such as precomputation of points.

By addressing these areas, we aim to significantly enhance MSM computation speed using Metal.

# Acceptance Criteria
1. The entire MSM process is implemented on the GPU, minimizing data conversion overhead.
2. Initialization of points and scalars for GPU computation is significantly faster.
3. The final implementation is compatible with Arkworks and passes all relevant tests.

vivianjeng:
move to https://github.com/zkmopro/gpu-acceleration/issues/1

PR #187, opened 2024-06-28
Feat/window-wise msm with precomputation acceleration on CPU

moven0831:
# Summary
This PR experiments with accelerating Arkworks MSM on the CPU using precomputation techniques. The result shows that while window-wise MSM with precomputation provides a slight speedup, the increased I/O and serialize/deserialize overhead are not worth it.

Check on this report for more detail: https://hackmd.io/v8WPAG8-RsCANrbv_seSXw

# Status
- Since the goal is to speed up MSM computation, we tried both GPU and CPU ways to accelerate it. The GPU way is recorded in this issue https://github.com/zkmopro/mopro/issues/153.

# This PR
Implement the generation for precomputed point and related serde methods. Adapted the Arkworks MSM algorithm to use the precomputed points.

# benchmark result (on MacBook with M3 chip)
> similar to Arkworks MSM, sometimes are slightly faster but not significant 

|         msm time (ms) / instance size        |  2^10 |   2^16  |   2^18   |   2^20   |    2^22   |    2^24   |
|:--------------------------------------------:|:-----:|:-------:|:--------:|:--------:|:---------:|:---------:|
|                 Arkworks msm                 |  2.68 |  67.96  |  245.38  |  912.32  |  3641.81  |  14254.36 |
| Window-wise MSM with precomputation <br> (on CPU) |  2.13 |  65.83  |  234.32  |  915.91  |     x     |     x     |

The relationship among acceleration speed, size of precomputation points and time for precomputation is as follow:
  - 2^10: precompute_factor=20 (i.e. precomputed point size increase 20 times)
  - 2^16: precompute_factor=20
  - 2^18: precompute_factor=19
  - 2^20: precompute_factor=17 (Size of precomputed points is around 570 MB)


vivianjeng (approved):
It can be a different feature flag to control different curves
so that you don't need to comment out `trapdoortech_zprize_msm` and `halo2curve_msm`

LGTM

PR #195, opened 2024-07-03
doc: add new benchmark result

vivianjeng:
- add benchmark result in different circuits


vimwitch (approved):
lgtm, cool that w2c2 is so much faster than wasmer

PR #199, opened 2024-07-03
feat: build witness in background thread

### LINKED ISSUE #198: Load zkey and build witness in parallel (circom)
# Problem
Loading the zkey and building the witness do not depend on each other. These operations can both take relatively large amounts of time.

## Details
We should load the zkey and build the witness in parallel before making the proof.

vimwitch:
Closes #198


vivianjeng (approved):
Awesome!
for me total keccak256 `generate_circom_proof` time on Mac is `1.3296585`s
`ark_proof` time is `1.141613084`s
the diff `0.188`s includes both `0.139`s loading zkey and `0.173`s witness generation

PR #207, opened 2024-07-09
feat: bls12-381 support

### LINKED ISSUE #203: Add support for bls381
# Problem
Mopro only supports circom proofs over the alt_bn128 curve. We rely on `circom-compat` to read the zkey. This package returns a `ProvingKey<Bn254>` structure from `read_zkey`.

## Details
We should add support for the bls381 curve to mopro as it will be added to ethereum as a precompile in the next hard fork. We'll need test zkeys/wasms built for the bls381 curve.

The zkey binary file includes information about what curve the key is built for. We should automatically use this to change what curve we're making proofs on.

## Acceptance criteria
Make a circom proof over the bls381 curve.

vimwitch:
Closes #203 

This PR implements zkey loading, proving, and verifying over a generic curve. We automatically detect the curve from the zkey and use the correct implementation, or error if using an unrecognized curve. Supported curves:
- `alt_bn128`
- `bls12_381`

We should check binary sizes when adding additional curves. It might be worth conditionally compiling the curves.

TODO:
- [x] add test bls zkey
- [x] load groth header with manual Fq/Fr parsing, no field type


vimwitch:
This PR is ready for review. I have wrote a poseidon implementation for bls and wanted to use that for the tests here but it's making incorrect proofs. I documented the issue in #208 and added a different circuit for testing bls.

vivianjeng (approved):
Good PR!
I tried on iOS simulator and it works for me
I think we can add some e2e tests
and update bindings tests
(I can help with these)

PR #211, opened 2024-07-09
chore: remove ark-circom

vivianjeng:
- remove `ark-circom` dependency

| build time | with ark-circom | without ark-circom  |
| -- | -- | -- |
| debug | 1m 07s | 42.80s |
| release |4m 00s | 3m 48s|


vimwitch:
I think instead of moving `ark-circom` into `mopro-ffi`, we should try to move logic out to `ark-circom`. The `zkey.rs` and `zkey_header.rs` could both go out to `ark-circom`. We can also try to push changes from our `ark-circom` to `arkworks-rs/circom-compat`. Specifically the zkey loading and native witness gen.

vivianjeng:
I see. That makes sense.
I will do it in another PR

PR #213, opened 2024-07-10
chore: move zkey to circom-compat

vivianjeng:
- move zkey logic to zkmopro/circom-compat
  - PR: https://github.com/zkmopro/circom-compat/pull/3


mopro-ffi/src/circom/mod.rs
```diff
@@ -67,26 +62,14 @@ pub fn generate_circom_proof_wtns(
...
         return prove(proving_key, matrices, full_assignment);
     } else if header_reader.r == BigUint::from(ark_bls12_381::Fr::MODULUS) {
-        let mut binfile = BinFile::<_, Bls12_381>::new(&mut reader)
-            .map_err(|e| MoproError::CircomError(e.to_string()))?;
-        let proving_key = binfile
-            .proving_key()
-            .map_err(|e| MoproError::CircomError(e.to_string()))?;
-        let matrices = binfile
-            .matrices()
+        let (proving_key, matrices) = read_zkey::<_, Bls12_381>(&mut reader)
```
vimwitch, line 72: I think we should keep these functions separate. `read_zkey` loads both the proving key and the matrices. In the verifier we only want to load the proving key, not the matrices. That's why i originally separated it to two functions.

vimwitch (changes requested):
we could merge as is, but it would be good to fix this first

mopro-ffi/src/circom/mod.rs
```diff
@@ -150,18 +135,12 @@ pub fn verify_circom_proof(
     let file = File::open(&zkey_path).map_err(|e| MoproError::CircomError(e.to_string()))?;
     let mut reader = std::io::BufReader::new(file);
     if header_reader.r == BigUint::from(ark_bn254::Fr::MODULUS) {
-        let mut binfile = BinFile::<_, Bn254>::new(&mut reader)
-            .map_err(|e| MoproError::CircomError(e.to_string()))?;
-        let proving_key = binfile
-            .proving_key()
+        let proving_key = read_proving_key::<_, Bn254>(&mut reader)
```
vivianjeng, line 138: @vimwitch I use a `read_proving_key` in circom-compat
I think it is more readable function (than like `BinFile`)

vimwitch: I agree, looks good

vimwitch (approved):

PR #214, opened 2024-07-10
Halo2 circuit in mopro (redesigned)

ElusAegis:
Proposition on how we could add Halo2 Circuits to redesigned Mopro.

### Overview
I attempted to be as consistent as possible to `Circom` implementation, resulting in defining two types `Halo2ProveFn` and `Halo2VerifyFn` which we expect the user to implement locally and then use the `key_halo2_circuit_map` as a way to map between different circuits and their implementations of prove and verify functions. 

_Note that as Halo2 is a library, there is **very little** shared components, hence `Mopro` only defines these two types as well as uses `key_halo2_circuit_map` to invoke correct function at run time. _ 

### Halo2 Module in `test-e2e` 
I have also added a `halo2` module in the `test-e2e` crate. It does not contain a halo2 circuit logic (I have extracted it into another repository as discussed before - https://github.com/ElusAegis/halo2-fibonacci-sample.git, which can be forked into Mopro organisation). What it has is logic to interface with the circuit, implementing logic to parse and prove the values. 

As each circuit can use its own types, I did not see a way to abstract it away without enforcing strict user limitations. So `halo2` is an example which users would need to reproduce. (Halo2 requires quite a bit of background knowledge).

### Halo2/Circom Features 

Currently, all users must use both Halo2 and Circom (define `key_halo2_circuit_map` and `zkey_witness_map`.  I have given a thought to how we could be able to disable and enable each of them, however with current design that does not seem to be possible unless we add something like `halo2_placeholder!()` which users need to specify if they do not use Halo2. This does not seem ideal UX.

Hence @vimwitch @vivianjeng , perhaps you have some views on how we could be able to do this? We may need to do something different than the current approach of making a user define a particular function.


ElusAegis:
@vimwitch @vivianjeng I have finished the PR. Can you please have a look. 


Key changes were to `app` macros, to enable to have different adapters.

I have it now recursively call `halo2_app` and `circom_app`, which each have a fallback implementation when their respective feature is not implemented.

I have also played with how the `test-e2e/lib.rs` would look like, considering we would need conditionally compile it based on the flags, and I ended up doing it using macros, which are conditionally expanded, thus avoiding to write  `#[cfg(feature = ...)]` before each line.


I have added tests for iOS, however I have not updated Android as I am still unable to compile it.

test-e2e/src/halo2/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        .ok_or(MoproError::Halo2Error(
+            "Failed to get `out` value".to_string(),
+        ))?
+        .clone();
+
+    let mut transcript = TranscriptWriterBuffer::<_, G1Affine, _>::init(Vec::new());
+
+    let public_input = vec![a, b, out];
+
+    create_proof::<KZGCommitmentScheme<Bn256>, ProverGWC<_>, _, _, Blake2bWrite<_, _, _>, _>(
```
vimwitch, line 110: I think this logic should also be in `halo2_fibonacci`, specifically the generics about commitment scheme and field.

vimwitch: Ideally the code here is just `halo2_fibonacci::prove(srs_path, proving_key, inputs)`

test-e2e/src/lib.rs
```diff
@@ -1,20 +1,50 @@
...
-rust_witness::witness!(multiplier2);
-rust_witness::witness!(multiplier2bls);
-rust_witness::witness!(keccak256256test);
-rust_witness::witness!(hashbenchbls);
+// First, configure the Mopro FFI library
+mopro_ffi::app!();
 
-app!();
+/// Macro containing the Circom example
+#[cfg(feature = "circom")]
```
vimwitch, line 7: I don't think we should do conditional compilation in this crate. test-e2e should test all proving systems.

in the test apps we should eventually add more screens to handle generating different types of proofs

test-e2e/src/halo2/mod.rs
```diff
@@ -0,0 +1,218 @@
+use std::collections::HashMap;
```
vimwitch, line 1: I don't think `test-e2e` should have more than 1 module (`lib.rs`). All other logic should be delegated to either `mopro-ffi`, or a custom prover (in this case `halo2_fibonacci`).

test-e2e/src/halo2/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        &[&[inputs.0.as_ref()]],
+        &mut transcript,
+    )
+    .map_err(|_| MoproError::Halo2Error("Failed to verify the proof".to_string()))?;
+
+    Ok(true)
+}
+
+#[cfg(test)]
+mod tests {
```
vimwitch, line 145: This isn't testing `mopro-ffi`, these tests should be in `halo2-fibonacci`

test-e2e/src/halo2/serialisation.rs
```diff
@@ -2,28 +2,32 @@ use std::collections::HashMap;
 use std::fmt;
```
vimwitch, line 2: also not really `test-e2e`'s concern

test-e2e/src/lib.rs
vimwitch: Removing conditional compilation should make this easier to understand. Macros are scary

test-e2e/tests/bindings/test_circom_keccak.swift
```diff
@@ -73,6 +73,8 @@ do {
 
 } catch let error as MoproError {
   print("MoproError: \(error)")
+  throw error
```
vimwitch, line 76: good catch

mopro-ffi/src/circom/mod.rs
```diff
@@ -29,6 +29,75 @@ use color_eyre::Result;
...
+
+/// Set the circuits that can be proven by the mopro library
+/// Provide the circuits that you want to be able to generate proofs for
+/// as a list of pairs of the form `zkey`, `wtns_fn`
+/// Where `zkey` is the name of the zkey file
+/// and `wtns_fn` is the function that generates the witness for the circuit
+/// NOTE: YOU CAN ONLY USE THIS MACROS ONCE IN YOUR CODEBASE, IN THE SAME MODULE AS `app!()`
+/// NOTE: TO USE THIS MACRO, YOU MUST HAVE THE `mopro-ffi/circom` FEATURE ENABLED
+#[macro_export]
+macro_rules! set_circom_circuits {
```
vimwitch, line 86: Ok so we're using `set_circom_circuits!` to generate a function in the same scope as the above macro. The once-cell above invokes this function to set it's value.

I think we should remove this macro and the once-cell. Users should author the mapping function in rust like before.

Using macros makes it less clear _what_ is happening behind the scenes. It also introduces a non-standard syntax. e.g. even if i know rust i have to figure out this macro.

mopro-ffi/src/halo2/mod.rs
```diff
@@ -1,69 +1,172 @@
...
-    // Ok(is_valid)
+/// Proving Circuits are Halo2 Circuits that can generate proofs
+/// Provide the circuits that you want to be able to generate proofs for
+/// as a list of pairs of the form `circuit_proving_key`, `prove_fn`
+/// Where `circuit_proving_key` is the name of the proving key file
+/// and `prove_fn` is the function that generates the proof available locally
+/// NOTE: YOU CAN ONLY USE THIS MACROS ONCE IN YOUR CODEBASE, IN THE SAME MODULE AS `app!()`
+/// NOTE: TO USE THIS MACRO, YOU MUST HAVE THE `mopro-ffi/halo2` FEATURE ENABLED
+#[macro_export]
+macro_rules! set_halo2_proving_circuits {
```
vimwitch, line 65: Same here, remove this macro and just let the user write a normal function with a certain name mapping inputs to proof functions.

mopro-ffi/src/halo2/mod.rs
```diff
@@ -1,69 +1,172 @@
...
-    assert!(is_valid);
+/// Verifying Circuits are Halo2 Circuits that can verify proofs
+/// Provide the circuits that you want to be able to verify proofs for
+/// as a list of pairs of the form `circuit_verifying_key`, `verify_fn`
+/// Where `circuit_verifying_key` is the name of the verifying key file
+/// and `verify_fn` is the function that verifies the proof available locally
+/// NOTE: YOU CAN ONLY USE THIS MACROS ONCE IN YOUR CODEBASE, IN THE SAME MODULE AS `app!()`
+/// NOTE: TO USE THIS MACRO, YOU MUST HAVE THE `mopro-ffi/halo2` FEATURE ENABLED
+#[macro_export]
+macro_rules! set_halo2_verifying_circuits {
```
vimwitch, line 88: Let's try to combine the prove/verify mapping. It should be possible to return a tuple of functions.

test-e2e/src/lib.rs
```diff
@@ -1,20 +1,50 @@
-use mopro_ffi::{app, WtnsFn};
+extern crate core;
 
-rust_witness::witness!(multiplier2);
-rust_witness::witness!(multiplier2bls);
-rust_witness::witness!(keccak256256test);
-rust_witness::witness!(hashbenchbls);
+// First, configure the Mopro FFI library
+mopro_ffi::app!();
```
vimwitch, line 4: Ok yes, if we remove the `set_` macros we end up with just this as a macro invocation.

vimwitch (changes requested):
Definitely moving in the right direction. Just need to pull some logic out of `test-e2e` and get rid of a couple macros in favor of written functions.

mopro-ffi/src/circom/mod.rs
ElusAegis: Personally (and this is definitely a biased opinion) I though that this would be much more consistent to ask the users to invoke two macros, as opposed to asking to invoking a macro and then implementing a function, which at first does not seem to have clear relevance. 

Especially as rust does not have function overloads, I thought macros was a standard approach. 

As for complexity - macro is a simple solution, however, indeed, it only covers for a function and can be replaced by a function if the user wants.

My take is that if we want a user just use our library and simplify it as much as possible, macro abstracts all of the complexity, while allowing to also take a more complex approach.

I also made sure to add docs to macros (as well as will improve them to mention that a user can do it themselves).

ElusAegis:
@vimwitch thank you for the review, I agree with all of it besides how we show the setup to users. Will get to improving the

test-e2e/src/halo2/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct SerializableProof(pub Vec<u8>);
+
+#[derive(Clone, Debug)]
+pub struct SerializablePublicInputs(pub Vec<Fr>);
+
+pub(crate) struct FibonacciMoproCircuit;
+
+impl FibonacciMoproCircuit {
+    pub(crate) fn prove(
```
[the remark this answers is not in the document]
ElusAegis: That's a topic I have alluded in the introduction to this PR. Halo2 has a lot of logic unique to how a circuit is launched. And as this is configurable, it is not usually placed into the circuit library. 

We can put it, however users will need to manually do it usually, which means that the sample would require users to likely also explore 'fibonacci' to understand how it works

.github/workflows/build-and-test.yml
```diff
@@ -1,190 +1,195 @@
 name: Build and Test
 
 on:
-    push:
```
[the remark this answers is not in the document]
ElusAegis: My bad - my text editor was quite aggressive. Reverted it to the original

test-e2e/src/lib.rs
ElusAegis: Cleaned it up, agree that this is a better way, considering test-e2e should test all proving systems.

mopro-ffi/src/circom/mod.rs
ElusAegis: I will get rid of the `one-cell` as well as the static variable, this was made redundant by the split macro approach.

test-e2e/src/halo2/mod.rs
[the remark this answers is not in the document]
ElusAegis: I have moved everything to the `halo2_fibonacci`.

mopro-ffi/src/circom/mod.rs
ElusAegis: I have significantly changed the underlying of the macros, so that we no longer use `one-cell` and instead they generate a match style function, similar to what was originally.

This effectively provides two ways for action for users:
They can simply use a macro (ideal for those who do not know how rust works)
```rust

    ("multiplier2_final.zkey", multiplier2_witness),
    ("multiplier2_bls_final.zkey", multiplier2bls_witness),
    ("keccak256_256_test_final.zkey", keccak256256test_witness),
}
```rust
fn get_circom_wtns_fn(circuit: &str) -> Result<mopro_ffi::WtnsFn, mopro_ffi::MoproError> {
   match circuit {
      "circuit1.zkey" => Ok(circuit1_witness_fn),
     _ => Err(mopro_ffi::MoproError::CircomError(format!("Unknown ZKEY: {}", circuit).to_string()))
  }
}
```

The doc comment provides a guide to these both options, though I personally think that implementing a function is more complex than using a macro, especially as we already use a macro to initiate `mopro_ffi`.

Furthermore, the function requires error handling, arguably out of scope for the users and only adds future complexity.

mopro-ffi/src/halo2/mod.rs
ElusAegis: Same thoughts as above

mopro-ffi/src/halo2/mod.rs
ElusAegis: I did combine them under the same macro (the simple approach):
```rust
mopro_ffi::set_halo2_circuits! {
    ("fibonacci_pk", halo2_fibonacci::prove, "fibonacci_vk", halo2_fibonacci::verify)
}
```

However, under the hood they are still different functions, to serve the advanced users who might want to have more tailored logic:

 As effectively a user can select if they want to have a halo2 functionality for only verifying or also proving. And that implies that they should have an option to only specify the circuit they want to be able to verify, while avoiding adding any proving circuits. By merging it all into a single function that removes the flexibility, which will likely be very important for the users.

ElusAegis:
There are further 3 samples which I have created to show how Halo2 would work with Mopro:
	- RSA Full Mopro example  (https://github.com/ElusAegis/halo2-rsa-mopro.git)
        -  Keccak256 Mopro Halo2 Circuit (with use example in mopro/test-e2e) https://github.com/ElusAegis/halo2-keccak-stable.git
        -  Fibonacci Mopro Halo2 Circuit (with use example in mopro/test-e2e) https://github.com/ElusAegis/halo2-fibonacci-sample.git

ElusAegis:
## Benchmarking:

**[Keccak256](https://github.com/ElusAegis/halo2-keccak-stable.git):**
1.  _M1 Pro iPhone simulator:_ `116s` to prove and `1.4s` to verify 
2. _iPhone 15 Pro:_ `200s` prove time, `1.1s` verify time

**[RSA](https://github.com/ElusAegis/halo2-rsa-mopro.git):**
1. _M1 Pro iPhone simulator:_` ~450s` proving , `~1s` verifying time (without proving/verifying key generation)
2. _iPhone 15 Pro:_  Crashes due to memory limit exceeded (needs around 5GB RAM, with iPhone limiting to 3GB)

mopro-ffi/src/circom/mod.rs
vimwitch: I like the change in the rust macro. Providing options as `(str, fn)` tuples is much more clear than just a list.

Having both options is probably the best solution 👍🏻

test-e2e/Cargo.toml
```diff
@@ -13,19 +13,33 @@ name = "ios"
...
+# Circom dependencies
+rust-witness = { git = "https://github.com/vimwitch/rust-witness.git" }
+num-bigint = { version = "0.4.0" }
+
+# Halo2 dependencies
+halo2-fibonacci = { git = "https://github.com/ElusAegis/halo2-fibonacci-sample.git" }
+halo2-keccak-256 = { git = "https://github.com/ElusAegis/halo2-keccak-stable.git" }
+
+bincode = "1.3.3"
+halo2curves = { git = "https://github.com/privacy-scaling-explorations/halo2curves", tag = "0.3.2" }
```
vimwitch, line 29: Can we get rid of some (most?) of these new deps?

test-vectors/halo2/fibonacci_pk
vimwitch, line 1: Can we put a file extension of these? Even just `.bin` to indicate that they are binary files.

vimwitch (commented):
Good work, this is just about ready to merge

test-vectors/halo2/fibonacci_pk
ElusAegis: Pushed the changes, hopefully the CI were not broken

vimwitch (approved):

vivianjeng (approved):
Looks good to me!
Some places I think it can be simplified as I commented
and some comment messages can be written in docs (can be in the following PRs)

.github/workflows/build-and-test.yml
```diff
@@ -71,7 +61,23 @@ jobs:
...
+              uses: actions-rs/toolchain@v1
+              with:
+                  toolchain: stable
+                  override: true
+            - name: Run e2e halo2 tests
+              # TODO: Fix this custom jar thing
+              run: |
+                  cd test-e2e
+                  curl -L https://repo1.maven.org/maven2/net/java/dev/jna/jna/5.13.0/jna-5.13.0.jar -o jna-5.13.0.jar
+                  CLASSPATH=jna-5.13.0.jar cargo test --test halo2 -- --nocapture
```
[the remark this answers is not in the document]
ElusAegis: Removed the `test-e2e-halo2`, and merged the test task step `test-e2e` of circom , which I have just renamed to test-e2e

mopro-ffi/src/halo2/mod.rs
```diff
@@ -1,69 +1,163 @@
...
+            let result = verify_halo2_proof(
+                SRS_KEY_PATH.to_string(),
+                VERIFYING_KEY_PATH.to_string(),
+                proof_result.proof,
+                proof_result.inputs,
+            );
+            assert!(result.is_ok());
+        } else {
+            panic!("Failed to generate proof")
+        }
```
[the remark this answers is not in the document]
ElusAegis: Done

ElusAegis:
@vivianjeng thank you for the review, I have implemented the suggestions. However, after I have merged with the main and updated Cargo.lock it no longer runs the  `iOS-app-simulator` successfully. Yet it does on my local machine with a clean start. 

I do remember we had some problems with non-determinism, could it be one of them? Or was there something in the recent PR that was merged that could have impacted it? I looked at it and nothing seems to be able to break the process at that point.

vimwitch:
The iOS simulator on github has intermittent failures 😭. Usually if you read the log it will say something like "timed out" or similar.

PR #219, opened 2024-07-18
doc: versioned doc

vivianjeng:
- set latest version to ~`0.0.2`~ `0.1.0`


vimwitch:
What is the difference between 0.0.1 and 0.0.2?

vimwitch (approved):

PR #221, opened 2024-07-22
Feat/integrate gpu acceleration

FoodChain1028:
Integrating msm functions in `gpu-acceleration` into `mopro-ffi`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
- "gimli",
+ "gimli 0.28.1",
 ]
 
 [[package]]
@@ -16,6 +16,17 @@ name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
+name = "ahash"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
+dependencies = [
+ "getrandom",
+ "once_cell",
+ "version_check",
+]
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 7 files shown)


vivianjeng:
I am thinking if we can make the mopro repo as simplified as this


and we just import the accelerated crate in Cargo.toml
e.g. https://github.com/zkmopro/mopro/blob/27787bbe7da0e4c2cf63455ab20305b15cc2e602/mopro-ffi/Cargo.toml#L83
or like we update the `circom-compat`: https://github.com/zkmopro/circom-compat/tree/wasm-delete

FoodChain1028:
@vivianjeng I see, so should we modify the functions into another `macro` like this? I am not sure what you mean sorry :(

https://github.com/zkmopro/mopro/blob/27787bbe7da0e4c2cf63455ab20305b15cc2e602/mopro-ffi/src/lib.rs#L159-L161

moven0831:
> I am thinking if we can make the mopro repo as simplified as this 
> 
> and we just import the accelerated crate in Cargo.toml e.g.
> 
> https://github.com/zkmopro/mopro/blob/27787bbe7da0e4c2cf63455ab20305b15cc2e602/mopro-ffi/Cargo.toml#L83
> 
> 
> or like we update the `circom-compat`: https://github.com/zkmopro/circom-compat/tree/wasm-delete

Now the structure is
(1) we import `gpu-acceleration/mopro-msm` as an optional crate and add it inside the `gpu-acceleration` feature
```toml
mopro-msm = { git = "https://github.com/zkmopro/gpu-acceleration.git", optional = true }
```
(2) extend corresponding functions and variables in `mopro-ffi/src/lib.rs` and `mopro-ffi/src/mopro.udl` for uniffi purpose
(3) generate new SwiftBindings

In which part do you think we could simplify further? We will try our best to find a minimized way to import the gpu-acceleration crate.

vivianjeng:
if it can accelerate circom, you can create a PR in `zkmopro/circom-compat` and replace the msm part. Or if it can accelerate halo2, you can replace like `halo2-fibonacci` with a new crate, e.g. `gpu-halo2-fibonacci`.
and once `zkmopro/circom-compat` or `gpu-halo2-fibonacci` is updated, you create another PR to specify which branch or version is used.

My idea is to keep the repo only binds with the `generateProof` and `verifyProof` functions
(but the msm is more like inside of `circom-compat` or `halo2`)

moven0831:
I see. Does it make sense to you @vivianjeng that we first close the PR and start working on integrating the `gpu-acceleration` crate into either `zkmopro/circom-compat` or `halo2`? After that, we can create another PR to specify which branch or version we used for the acceleration, as you've mentioned.

By the way, I found that we need to fork `ark-groth16` and integrate our MSM logic inside [prover.rs](https://github.com/arkworks-rs/groth16/blob/master/src/prover.rs#L53-L132). Then, we can use the modified `groth16` in `zkmopro/circom-compat`.

cc @FoodChain1028

PR #222, opened 2024-07-23
v0.1.0

vivianjeng:


README.md
```diff
@@ -51,11 +51,11 @@ default = ["mopro-ffi/circom"]
 [dependencies]
 mopro-ffi = { git = "https://github.com/zkmopro/mopro.git" }
 uniffi = { version = "0.28", features = ["cli"] }
-rust-witness = { git = "https://github.com/vimwitch/rust-witness.git" }
+rust-witness = "=0.1.0"
```
chancehudson, line 54: Why is this pinned to `=0.1.0`? I think it could be good to do just `0.1.0` so we can push fixes to rust-witness without needing to update mopro

mopro-ffi/README.md
```diff
@@ -1,8 +1,27 @@
...
+## Getting started
+
+- Make sure you've installed the [prerequisites](https://zkmopro.org/docs/prerequisites).
+- Getting started with this [tutorial](https://zkmopro.org/docs/getting-started/rust-setup).
+
+## Run tests
+
+- circom
+  ```sh
+  cargo test --features circom --no-default-features
```
chancehudson, line 14: I think we don't need `--no-default-features` in these calls. Could be good to remove that flag and simplify these commands.

PR #228, opened 2024-08-01
refactor: cleanup error handling

### LINKED ISSUE #220: Use `anyhow` crate for error handling
# Problem
Right now we use custom error type. In the code we unwrap errors and map them into the custom error type like [here](https://github.com/zkmopro/mopro/blob/5525c77b92db976cab2d13e58d7ed59bebd7767a/mopro-ffi/src/circom/mod.rs#L151). This is inconvenient and leads to lots of extra code.

## Details
Use the [`anyhow`](https://crates.io/crates/anyhow) crate to propagate errors outward. Unwrap the error and rewrap into a `MoproError` only at the top level

## Acceptance criteria
All `MoproError` mappings are removed, except for at the top level.

chancehudson:
This PR refactors the internal functions to use `anyhow` instead of `MoproError`. The error is wrapped in `MoproError` in the udl functions only to avoid many `map_err` calls in core logic.

Closes #220

### DIFF
--- a/mopro-ffi/src/circom/mod.rs
+++ b/mopro-ffi/src/circom/mod.rs
@@ -250,7 +250,7 @@ mod tests {
     use std::str::FromStr;
 
     use crate::circom::{generate_circom_proof_wtns, serialization, verify_circom_proof, WtnsFn};
-    use crate::{GenerateProofResult, MoproError};
+    use crate::GenerateProofResult;
     use anyhow::bail;
     use anyhow::Result;
     use ark_bls12_381::Bls12_381;
@@ -352,7 +352,7 @@ mod tests {
     }
 
     #[test]
-    fn test_prove() -> Result<(), MoproError> {
+    fn test_prove() -> Result<()> {
         // Create a new MoproCircom instance
         let zkey_path = "../test-vectors/circom/multiplier2_final.zkey".to_string();
 
@@ -399,7 +399,7 @@ mod tests {
     }
 
     #[test]
-    fn test_prove_keccak() -> Result<(), MoproError> {
+    fn test_prove_keccak() -> Result<()> {
         // Create a new MoproCircom instance
         let zkey_path = "../test-vectors/circom/keccak256_256_test_final.zkey".to_string();
         // Prepare inputs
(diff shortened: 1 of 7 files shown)


mopro-ffi/src/circom/mod.rs
```diff
@@ -142,33 +145,23 @@ pub fn generate_circom_proof_wtns(
...
     // check the prime in the header
     // println!("{} {} {}", header.q, header.n8q, ark_bls12_381::Fq::MODULUS);
     if header_reader.r == BigUint::from(ark_bn254::Fr::MODULUS) {
-        let (proving_key, matrices) = read_zkey::<_, Bn254>(&mut reader)
-            .map_err(|e| MoproError::CircomError(e.to_string()))?;
-        let full_assignment = witness_thread
-            .join()
-            .map_err(|_e| MoproError::CircomError("Failed to generate witness".to_string()))?;
+        let (proving_key, matrices) = read_zkey::<_, Bn254>(&mut reader)?;
+        let full_assignment = witness_thread.join().unwrap();
```
vivianjeng, line 154: will the error messages here be caught or shown? if there are errors.
e.g. previously there are errors `e.to_string()` or `"Failed to generate witness"`

mopro-ffi/src/circom/mod.rs
```diff
@@ -212,34 +205,30 @@ pub fn verify_circom_proof(
     zkey_path: String,
     proof: Vec<u8>,
     public_input: Vec<u8>,
-) -> Result<bool, MoproError> {
+) -> Result<bool> {
     let mut header_reader = ZkeyHeaderReader::new(&zkey_path);
     header_reader.read();
-    let file = File::open(&zkey_path).map_err(|e| MoproError::CircomError(e.to_string()))?;
+    let file = File::open(&zkey_path)?;
```
vivianjeng, line 211: same here
and the following statement replacing `e.to_string()` to `?`

mopro-ffi/src/circom/mod.rs
chancehudson: It would cause a panic. I caught it and provided a nicer error message.

vivianjeng (approved):
LGTM

Issue #234, opened 2024-08-24
Mopro Cli v0.1.0


vivianjeng:
# Problem
A cli like `circom`, foundry, `create-react-app`... is helpful for developers getting started with a new project.

## Details
- create a `mopro-cli` directory including the following commands
1. `mopro init`: create an example rust project following the tutorial: [Rust Setup](https://zkmopro.org/docs/getting-started/rust-setup)
   - (Optional) Let user input a project name
   - (Optional) Let user specify proving systems
   - (Optional) Let user input circuits path
2. `mopro build`: export `MoproiOSBindings`, `MoproAndroidBindings`
3. `mopro create`: create the mobile app itself
   - Options: `platforms`
       - `swift` (template: [iOS Setup](https://zkmopro.org/docs/getting-started/ios-setup))
       - `kotlin` (template: [Android Setup](https://zkmopro.org/docs/getting-started/android-setup)
       - `react-native` (template: React Native Setup)

## Acceptance criteria
Make sure `mopro init`, `mopro build`, `mopro create` works and the apps work

## Next steps (optional)
- Documentation:
**Getting started** with the cli
And move the current setups to **Tutorial** section

oskarth:
For reference, here's the previous issue https://github.com/zkmopro/mopro/issues/77 that was addressed with `mopro-cli`

Documentation here https://zkmopro.org/docs/0.0.1/mopro-cli

Example usage here https://www.loom.com/share/6ff382b0497c47aea9d0ef8b6e790dd8

vivianjeng:
`mopro init` is done #235 
`mopro build` is opened #237

vivianjeng:
`mopro create` is done #245 (for iOS and Android)
TODO: react-native and flutter

PR #235, opened 2024-08-25
Cli for initializing a mopro project

vivianjeng:
- related: #234 
- `mopro init`: create a rust project through the [Rust Setup](https://zkmopro.org/docs/getting-started/rust-setup)
   ```sh
   (base) ➜  cli git:(cli) ✗ mopro init --help
   Initializes a new project

   Usage: mopro init [OPTIONS]

   Options:
         --adapter <ADAPTER>            Specify the adapter to use (e.g., 'circom', 'halo2' or 'circom,halo2').
         --project-name <PROJECT_NAME>  
     -h, --help                         Print help
   ```

- users can either specify `--adpater` or `--project-name`
   or choose adapters, project_name after promps
```sh
(base) ➜  cli git:(cli) ✗ mopro init
✔ Project name · mopro-example-app
✔ Pick the adapters you want to use (multiple selection with space) · circom
🚀 Project 'mopro-example-app' initialized successfully! 🎉

To get started, follow these steps:

1. Navigate to your project directory:
   cd mopro-example-app

2. Run the following commands to build and run the project:
   cargo run --bin ios
   cargo run --bin android

📚 To learn more about mopro, visit: https://zkmopro.org

Happy coding! 🚀
```

- `adapter` could be `circom`, `halo2`, or `circom,halo2`


cli/src/init.rs
```diff
@@ -0,0 +1,238 @@
...
+                if a.contains(adapter) {
+                    selection.push(i);
+                }
+            }
+            selection
+        }
+    };
+
+    if selection.is_empty() {
+        println!("\x1b[33mNo adapters selected. Use space to select an adapter.\x1b[0m");
```
chancehudson, line 41: I'd recommend using something like [this](https://crates.io/crates/console) or [this](https://crates.io/crates/yansi) to handle coloring/formatting terminal text.

cli/src/init.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        if let Some(cargo_toml_path) = project_dir.join("Cargo.toml").to_str() {
+            let _ = replace_project_name(cargo_toml_path, &project_name);
+            let _ = replace_features(
+                cargo_toml_path,
+                selection.iter().map(|&i| adapters[i]).collect(),
+            );
+            if selection.contains(&0) {
+                // circom is selected
+                let _ = circom_dependencies_template(cargo_toml_path);
```
chancehudson, line 63: Why are you assigning these? Why not just drop the value?

```rust
// e.g. just call without assignment
circom_dependencies_template(cargo_toml_path);
```

cli/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+    Init {
+        #[arg(
+            long,
+            help = "Specify the adapter to use (e.g., 'circom', 'halo2' or 'circom,halo2')."
+        )]
+        adapter: Option<String>,
+        #[arg(long)]
+        project_name: Option<String>,
+    },
+    // /// Prepare and build circuit artifacts
```
chancehudson, line 34: What are these comments for? Can they be removed?

cli/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+    match &cli.command {
+        // Commands::Deps {} => deps::install_deps(),
+        Commands::Init {
+            adapter,
+            project_name,
+        } => match init::init_project(adapter, project_name) {
+            Ok(_) => {}
+            Err(e) => println!("\x1b[1;31mFailed to initialize project {:?}\x1b[0m", e),
+        },
+        // Commands::Prepare { config } => prepare::prepare_circuit(config),
```
chancehudson, line 90: Same here

chancehudson:
I'm still opposed to adding a cli for managing mopro. By doing so we're absorbing the complexity of `cargo` and replacing it with out own complexity `mopro`. This _can_ be good in some situations, but i think in this case the gains are close enough to zero to make it not worth maintaining.

Basically, if people want to use mopro they'll do it with or without a cli. So IMO it only makes sense to improve the offering mopro itself (like adding react-native, more targets, etc).

That said, I'm not as involved in this project these days, and won't say no to a feature you think is correct. The implementation here looks good.

cli/src/init.rs
vivianjeng: there are warnings here
```sh
unused `Result` that must be used
this `Result` may be an `Err` variant, which should be handled
`#[warn(unused_must_use)]` on by defaultrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2FUsers%2Fzhengyawen%2FDocuments%2FGitHub%2Fmopro%2Fcli%2Fsrc%2Finit.rs)
init.rs(63, 17): use `let _ = ...` to ignore the resulting value: `let _ = `
```

but I can use `?` to void this warning as well

chancehudson: Ah yes, you want to either use the variable, or use `?`. The current implementation will silence any errors.

In the current implementation if `circom_dependencies_template` returns an error the program will continue as normal. You should use `?` or something like this:

```rust
if let Err(e) = circom_dependencies_template(cargo_toml_path) {
    println!("error in circom dependencies template {e}");
}
```

to unwrap the error

vivianjeng: It should be caught here
https://github.com/zkmopro/mopro/blob/da913a0ff3b199b33060535f344c473e9e285455/cli/src/main.rs#L88
if there is any error in this file (even with a `?`)

chancehudson: No, errors are only propagated if you use the `?` operator, or return the value.

So either of these would work:
```rust
return circom_dependencies_template(cargo_toml_path);
```
```rust
circom_dependencies_template(cargo_toml_path)?;
```

Bot not assigning a variable then not using it. [Here's an example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=511128252eb58b1c1b4265e500abbca9) of the problem.

cli/src/init.rs
vivianjeng: [3bbec3f](https://github.com/zkmopro/mopro/pull/235/commits/3bbec3fcbbacd7a329e6441a68257eac1a0bf193)

cli/src/init.rs
vivianjeng: [132638c](https://github.com/zkmopro/mopro/pull/235/commits/132638c70c3e17db08632bd62a1933efe05006b7)

cli/src/main.rs
vivianjeng: [132638c](https://github.com/zkmopro/mopro/pull/235/commits/132638c70c3e17db08632bd62a1933efe05006b7)

cli/src/main.rs
vivianjeng: [132638c](https://github.com/zkmopro/mopro/pull/235/commits/132638c70c3e17db08632bd62a1933efe05006b7)

vivianjeng:
My idea is like `npx create-create-app`
it could be very simple
(or at least we build a mopro circom example project)
and people can just `cd project` and open a simulator to run a prover

but I just found that the we can still keep these steps
(just too complex to do them in one step)

I think cli will be helpful for junior developers or for people who want to try new things
and for advanced (circuits) developers we can still keep how to setup in the documentation

yes I agree it will take much effort to maintain if there is any change in mopro
we can try to fix the version and make sure the cli works for every update (e.g. with a ci test) or every released version

cli/src/init.rs
```diff
@@ -0,0 +1,242 @@
...
+                circom_lib_template(lib_rs_path)?;
+            }
+            if selection.contains(&1) {
+                // halo2 is selected
+                halo2_lib_template(lib_rs_path)?;
+            }
+        }
+
+        // Print out the instructions
+        println!(
```
chancehudson, line 90: This isn't a necessary change, but you can do multiline strings by default. e.g.

```rust
let hello = "hello
world
more text
";
println!("{hello}");
```

cli/src/init.rs
```diff
@@ -0,0 +1,242 @@
...
+                }
+            }
+            if let Some(path_str) = parent.to_str() {
+                if path_str.contains("halo2") && !selection.contains(&1) {
+                    return Ok(());
+                }
+            }
+            if let Err(e) = fs::create_dir_all(parent) {
+                return Err(e);
+            }
```
chancehudson, line 133:
```suggestion
            fs.create_dir_all(parent)?;
```

chancehudson (approved):
lgtm

Issue #227, opened 2024-08-01
Folding schemes support


vivianjeng:
# Problem
Folding schemes are used by many projects and there are many mature tools: https://github.com/lurk-lab/awesome-folding?tab=readme-ov-file

## Details
Support one of the folding schemes (or many).

## Acceptance criteria
Generating a folding proof in iOS and/or Android and verify it.

vivianjeng:
@csiejimmyliu @june-in-exile @0xgtim will solve this issue

sifnoc:
Reference: https://github.com/privacy-scaling-explorations/sonobe

PR #240, opened 2024-09-21
NOVA Scotia folding scheme test done #227

csiejimmyliu:
- [Nova Scotia](https://github.com/nalinbhardwaj/Nova-Scotia) folding test in morpo ffi
- Using toy r1cs example of Fibonacci sequence


vivianjeng (commented):
It looks good! Thank you for your contribution
there are several points that can be changed
1. rename the circuit (`toy` -> e.g.`adder`)
2. remove temporary files `jna-5.13.0.jar`, `toy_js/` ( we just need the `r1cs` and `wasm`
3. try to run test with `cargo test --features nova_scotia` and include the CI script like
https://github.com/zkmopro/mopro/blob/b625093b2cc3f693640a56de769ac17dd2bd3eeb/.github/workflows/build-and-test.yml#L35-L36

mopro-ffi/Cargo.toml
```diff
@@ -71,6 +73,12 @@ ark-bls12-381 = { version = "0.4.0", optional = true }
 num-traits = { version = "0.2.0", optional = true }
 anyhow = "1.0.86"
 
+# Nova Scotia dependencies
+nova-scotia = "0.5.0"
+nova-snark = "0.23.0"
+serde_json = "1.0.85"
+pasta_curves = "0.5"
```
vivianjeng, line 80: these dependencies should be set `optional`
e.g.
```toml
nova-scotia = { version="0.5.0", optional = true}
```

csiejimmyliu: Testing.
I move those dependencies after [dev-dependencies].

vivianjeng: you still need to set it `optional` since it is in `nova_scotia` feature
the error is
```sh
`cargo metadata` exited with an error: error: failed to load manifest for workspace member `/home/runner/work/mopro/mopro/mopro-ffi`
referenced by workspace at `/home/runner/work/mopro/mopro/Cargo.toml`

Caused by:
  failed to parse manifest at `/home/runner/work/mopro/mopro/mopro-ffi/Cargo.toml`

Caused by:
  feature `nova_scotia` includes `nova-scotia`, but `nova-scotia` is not an optional dependency
  A non-optional dependency of the same name is defined; consider adding `optional = true` to its definition.

...
```

I suppose all the nova-scotia circuits will use the features like `nova-scotia`?
(and even all the dependencies??)
it won't be effected by the circuit's difference
e.g. I use `fibonacci` circuit, and I use `poseidonHash` circuit, I will use the same `nova-scotia` (but different `r1cs` and `wasm`)

mopro-ffi/src/lib.rs
```diff
@@ -67,6 +72,30 @@ macro_rules! halo2_app {
...
+            start_public_input: Vec<F<G1>>,
+        ) -> Result<GenerateProofResult, MoproError> {
+            panic!("Nova Scotia is not enabled in this build. Please pass `nova-scotia` feature to `mopro-ffi` to enable Nova Scotia.")
+        }
+
+        fn verify_nova_scotia_proof(
+            recursive_snark: RecursiveSNARK<G1, G2, C1<G1>, C2<G2>>,
+            pp: &PublicParams<G1, G2, C1<G1>, C2<G2>>,
+            iteration_count: usize,
+            start_public_input: &[E1::Scalar],
```
vivianjeng, line 92: I found that there are some problems with these data types
it causes the CI error (refer to: https://github.com/zkmopro/mopro/actions/runs/11051448766/job/30701493136)

in mopro-ffi tests you import these dependencies [here](https://github.com/csiejimmyliu/mopro/blob/a7bbad4d781effe38ef2e0551ac4db2532f4d3c5/mopro-ffi/src/nova_scotia/mod.rs#L92)
but like `test-e2e` doesn't have these dependencies so it says like
```sh
use of undeclared type `E1`
```

The possible solution might be: 
1. Use serialization like circom [serialization.rs](https://github.com/zkmopro/mopro/blob/main/mopro-ffi/src/circom/serialization.rs). 
   To make the type the native rust types: `Vec`, `String`, `u8`, `std::collections::HashMap`,...
2. or import the types in `src/lib.rs` (not recommended, but could be fast to solve)
   Since we will convert the type in `mopro.udl`, and the built-in types are: https://mozilla.github.io/uniffi-rs/0.27/udl/builtin_types.html

vivianjeng:
and please use `cargo fmt --all` to format the code

.github/workflows/build-and-test.yml
```diff
@@ -46,6 +48,18 @@ jobs:
                   override: true
             - name: Run ffi circom tests
               run: cd mopro-ffi && cargo test --features circom --no-default-features
+    test-ffi-nova_scotia:
```
vivianjeng, line 51: can we align the usage of `-` and `_`
I prefer the feature flag be `nova-scotia` 
e.g. `--features nova-scotia`
(but in rust code some place needs to be `nova_scotia` then just keep them `nova_scotia`)

mopro-ffi/Cargo.toml
```diff
@@ -35,6 +35,12 @@ circom = [
     "num-traits",
     "byteorder",
 ]
+nova_scotia = [
+    "nova-scotia",
```
vivianjeng, line 39: this can be changed to
```toml
nova-scotia = [
    "dep:nova-scotia",
```
to activate the dependency with the same name

test-e2e/Cargo.toml
```diff
@@ -25,6 +25,11 @@ num-bigint = { version = "0.4.0" }
 halo2-fibonacci = { git = "https://github.com/ElusAegis/halo2-fibonacci-sample.git" }
 halo2-keccak-256 = { git = "https://github.com/ElusAegis/halo2-keccak-stable.git" }
 
+# Nova Scotia dependencies
+nova-scotia = { version = "0.5.0" }
+nova-snark = { version="0.23.0" }
+serde_json = { version="1.0.85" }
+
```
vivianjeng, line 32: we should remove this (for now)
since we don't activate the `nova-scotia` feature
(and it needs to work without these dependencies)

but once you update the type for `udl` (serialization)
it should be solved easily
it can be done in another PR

vivianjeng (commented):
Thank you for fixing the errors
Can you fix the feature's name to `nova-scotia`
Thank we can merge this PR 🙏🏻

csiejimmyliu:
feature flag changed
@vivianjeng thanks for guiding

vivianjeng:
Sorry we are not going to extend proving systems in our roadmap 🙏🏻
and since currently there is no project using folding scheme
we don't want to maintain the code which is not being used

Thank you for your contribution, I think it is still helpful for people who want to integrate folding schemes in mobile

If you are still interested in the integration like this, we want to support Noir in the future 
https://github.com/zkmopro/mopro/issues/353
Since it is more widely used nowadays

PR #244, opened 2024-10-15
feat: ashlang spartan proving

chancehudson:
This PR is partially blocked by microsoft/Spartan#75. This pr uses a fork of spartan that is more friendly to proof serialization. As a result it cannot use a crates.io version of ashlang (git dependencies are not allowed in public packages).

[done](https://github.com/chancehudson/ashlang/commit/696960a0c15db47170fdd1ff058682023d904b1b): generate params for ar1cs dynamically and compare to provided proof


chancehudson:
this doesn't add android tests because i don't have JNA stuff installed on my current computer. I also have not run this code on device

mopro-ffi/Cargo.toml
```diff
@@ -17,8 +17,9 @@ name = "uniffi-bindgen"
 path = "src/uniffi-bindgen.rs"
 
 [features]
-default = []
+default = ["ashlang"]
```
vivianjeng, line 20: can you revert this?

mopro-ffi/Cargo.toml
```diff
@@ -38,11 +39,18 @@ circom = [
...
 
 # Error handling
 thiserror = "=1.0.39"
 color-eyre = "=0.6.2"
 
+# ashlang deps
+#ashlang = { path = "../../ashlang/ashlang", optional = true, default-features = false, features = ["spartan-prover", "serde"] }
+ashlang = { git = "https://github.com/chancehudson/ashlang.git", rev = "696960a0c15db47170fdd1ff058682023d904b1b", optional = true, default-features = false, features = ["spartan-prover", "serde"] }
+#ashlang = { version = "0.1.0", optional = true, default-features = false, features = ["spartan-prover"] }
+#scalarff = { git = "https://github.com/chancehudson/scalarff.git", branch = "no-twenty-first", optional = true }
```
vivianjeng, line 52: this should be cleaned up as well

mopro-ffi/src/ashlang/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+) -> anyhow::Result<GenerateProofResult> {
+    let ir_source = fs::read_to_string(ar1cs_path)?;
+    // we pass an empty vec for public inputs because
+    // they are not supported in the ashlang spartan prover
+    // outputs are public and should be used instead
+    let proof = SpartanProver::prove_ir(&ir_source, vec![], secret_inputs.clone())?;
+
+    Ok(GenerateProofResult {
+        proof: bincode::serialize(&proof)?,
+        inputs: bincode::serialize(&secret_inputs)?,
```
vivianjeng, line 47: It is not used in the `verify()`?

mopro-ffi/src/ashlang/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+    })
+}
+
+/// Verifies a spartan proof from an ar1cs file
+pub fn verify(ar1cs_path: &str, proof: Vec<u8>) -> anyhow::Result<bool> {
+    let ir_source = fs::read_to_string(ar1cs_path)?;
+    let p = bincode::deserialize(&proof)?;
+    ashlang::SpartanProver::verify(&ir_source, p)
+    // Ok(true)
```
vivianjeng, line 56:
```suggestion
```

mopro-ffi/src/ashlang/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_ashlang_prove_verify() -> anyhow::Result<()> {
+        let ar1cs_path = "../test-vectors/ashlang/example.ar1cs".to_string();
```
vivianjeng, line 65: I am thinking about to use a more meaningful circuit
e.g. `multiplier2` so that we can compare with circom prover

vivianjeng (commented):
I think if each circuit only depends on `ar1cs`, we don't need the macro to setup circuits.
we can just replace `prove` with `generate_ashlang_spartan_proof`?

and it works well for my iPhone 👍🏻

vivianjeng:
@chancehudson can you also include CI tests? Thank you

vivianjeng:
add tests here: https://github.com/zkmopro/mopro/pull/246

chancehudson:
> I think if each circuit only depends on ar1cs, we don't need the macro to setup circuits.
> we can just replace prove with generate_ashlang_spartan_proof?

No, uniffi needs the app to define the functions which is what the macro does. Without it app developers will need to write a function with the correct name themselves.

mopro-ffi/src/ashlang/mod.rs
chancehudson: Yes, this is the shared `GenerateProofResult` object.

mopro-ffi/src/ashlang/mod.rs
chancehudson: What do you want to compare?

mopro-ffi/src/ashlang/mod.rs
vivianjeng: so is there a concept of public signal?

mopro-ffi/src/ashlang/mod.rs
vivianjeng: like why people choose from circom to spartan 
In this example, the multiplier takes 0.02s to prove but the `ashlang_spartan_prover` takes 1s to prove
it could be hard to convince people what is the advantage of using `ashlang_spartan_prover`

mopro-ffi/src/ashlang/mod.rs
chancehudson: Ah, these `inputs` are meant to be public? There is no concept of public signal in Spartan i think, but ashlang does have support for such a thing (TritonVM for example has public inputs and secret inputs).

mopro-ffi/src/ashlang/mod.rs
chancehudson: Ah, the purpose of this implementation is not to convince someone to use spartan over groth16. IMO benchmarks should happen independently/separately from this integration.

I don't care much about the spartan prover though, it's ECC based so given the choice I'd use the stark prover (though it doesn't support 32 bit architectures right now).

mopro-ffi/src/ashlang/mod.rs
vivianjeng: It seems mentioned the `public inputs` in the readme
https://github.com/microsoft/Spartan?tab=readme-ov-file
but you can do it in the future PR

mopro-ffi/src/ashlang/mod.rs
vivianjeng: I am thinking about the `example` circuit is neither meaningful (people won't know it proves for a hash or some operations) nor fast
like `multiplier2` and `fibonacci` is the most simple/fast circuit in circom and halo2
can we switch to a fast and meaningful circuit (e.g. `multiplier2`)?

vivianjeng: looking forward to a stark prover 😄

chancehudson: Public inputs and outputs are not implemented in the [ashlang spartan prover](https://github.com/chancehudson/ashlang/blob/main/ashlang/src/provers/spartan.rs) yet (PR's are certainly welcome though). I'm a bit short on bandwidth right now and need to switch back to the LaBRADOR implementation I'm working on. I'd recommend merging this and keeping it behind a feature flag/out of the docs until the public inputs/outputs are supported.

chancehudson: Essentially all that needs to happen is the r1cs format is transformed to match what the spartan prover expects. Public inputs and outputs are already supported by the language/compiler they just need to be bridged to the spartan prover.

PR #245, opened 2024-10-22
feat(cli): create ios/android template

vivianjeng:
#234


cli/src/build.rs
```diff
@@ -1,4 +1,7 @@
-use crate::style::{self, blue_bold, print_green_bold};
+use crate::{
```
chancehudson, line 1: In my rust files I put 1 import per line to prevent merge conflicts. I order the imports as `std` first, external crates second, and local modules third. A good example is [here](https://github.com/Neptune-Crypto/twenty-first/blob/master/twenty-first/src/amount/u32s.rs#L26)

cli/src/create.rs
```diff
@@ -0,0 +1,189 @@
...
+use dialoguer::{theme::ColorfulTheme, Select};
+use include_dir::{include_dir, Dir};
+
+use crate::print::{print_create_android_success_message, print_create_ios_success_message};
+use crate::style::{self};
+
+// TODO: add  "react-native", "flutter"
+const TEMPLATES: [&str; 2] = ["ios", "android"];
+
+pub fn create_project(arg_platform: &Option<String>) -> Result<(), Box<dyn Error>> {
```
chancehudson, line 12: You should consider using the [`anyhow`](https://docs.rs/anyhow/latest/anyhow/) crate for errors. Here you're returning a memory reference to the dynamic `Error` trait. The `anyhow` crate wraps that complexity in a trait compatible with all `std` api's.

```suggestion
pub fn create_project(arg_platform: &Option<String>) -> anyhow::Result<()> {
```

cli/src/build.rs
sifnoc: > In my rust files I put 1 import per line to prevent merge conflicts. I order the imports as `std` first, external crates second, and local modules third. A good example is [here](https://github.com/Neptune-Crypto/twenty-first/blob/master/twenty-first/src/amount/u32s.rs#L26)

I follow this rule as well. we could consider enabling [group_imports](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#group_imports) in **rustfmt** to help avoid conflicts and keep lines shorter.
It’s an unstable feature but might be worth trying.

vivianjeng: @sifnoc Yes currently I don't follow the order to import packages
We can do that in another PR 👍🏻

PR #247, opened 2024-10-22
Replacing halo2 example

sifnoc:


mopro-ffi/SwiftBindings/mopro.swift
```diff
@@ -21,21 +21,21 @@ private extension RustBuffer {
     }
 
     static func empty() -> RustBuffer {
-        RustBuffer(capacity: 0, len: 0, data: nil)
+        RustBuffer(capacity: 0, len:0, data: nil)
     }
 
     static func from(_ ptr: UnsafeBufferPointer<UInt8>) -> RustBuffer {
-        try! rustCall { ffi_mopro_bindings_rustbuffer_from_bytes(ForeignBytes(bufferPointer: ptr), $0) }
+        try! rustCall { ffi_mopro_ffi_rustbuffer_from_bytes(ForeignBytes(bufferPointer: ptr), $0) }
```
vivianjeng, line 28: please revert this file change 🙏🏻 (and also  `moproFFI.h`)
It would be confusing when using `./build_bindings.sh`
If there is no ffi changes, please just keep the previous one

> [!NOTE]  
> We need to change the `mopro-ffi/Cargo.toml` to
> ```toml
> [lib]
> name = "mopro_bindings"
> ```
> And the functions would be `ffi_mopro_bindings_{}`
> but later on, the `test-e2e` uses `mopro_ffi` lib name
> so it has to switch to `name = "mopro_ffi"`
> 
> The solution could be
> 1. just using `mopro_ffi` as the lib name
> 2. config lib name in [`uniffi.toml`](https://mozilla.github.io/uniffi-rs/latest/swift/configuration.html)

vivianjeng (commented):
The other tests work for me
Just need to revert the `mopro.swift` and `moproFFI.h`

mopro-ffi/SwiftBindings/mopro.swift
sifnoc: Reverted

test-e2e/expand.rs
```diff
@@ -0,0 +1,1994 @@
+#![feature(prelude_import)]
+#[prelude_import]
+use std::prelude::rust_2021::*;
+#[macro_use]
+extern crate std;
+extern crate core;
```
vivianjeng, line 6: not sure why there is the file?

vivianjeng (changes requested):
and please revert the `jna-5.13.0.jar` file
(actually we should use`.gitignore` to ignore the file changing)

test-e2e/expand.rs
sifnoc: My fault while rebasing branch. those two files should not be included. (I created the file for checking the macro result)
I removed them at [e3b5198](https://github.com/zkmopro/mopro/pull/247/commits/e3b51989d8ad4d07f7b4dfa154d23d01e42e2d2d)

vivianjeng (approved):
Thank you so much

PR #253, opened 2024-11-04
Support customized UDL file.

alxkzmn:
- Do not overwrite the existing UDL file;
- If UDL exists, generate bindings based on that file instead of copying.
- If UDL file does not exist, write the default UDL in the project `src` directory.


vivianjeng:
(forward the message from discord)
I found out the error is `no such module 'moproFFI'`
and it also happens with my mac

the temporarily broken one 
the error message should be like
```sh
Testing failed:
    The test runner encountered an error (Failed to prepare device 'Clone 1 of iPhone 15' for impending launch. (Underlying Error: Timed out trying to boot simulator after waiting 60.00s.))
Test session results, code coverage, and logs:
```

cli/src/template/init/build.rs
```diff
@@ -1,10 +1,15 @@
...
+    // to your app. We have pre-generated this file for you.
+    // Feel free to modify it to suit your needs.
+    let udl_path = Path::new("src/mopro.udl");
+    if !udl_path.exists() {
+        std::fs::write(udl_path, mopro_ffi::app_config::UDL).expect("Failed to write UDL");
+    }
     // Finally initialize uniffi and build the scaffolding into the
     // rust binary
-    uniffi::generate_scaffolding("./src/mopro.udl").unwrap();
+    uniffi::generate_scaffolding(udl_path).unwrap();
```
vivianjeng, line 14: it should be a string type
e.g. `udl_path.to_str().unwrap()`

mopro-ffi/src/app_config/android.rs
```diff
@@ -32,10 +30,35 @@ pub fn build() {
...
 
-    write_kotlin_bindings(&bindings_out);
+    let uniffi_bindgen_path = {
+        let mut bin_path = std::env::current_exe().unwrap(); //Users/~/mopro/target/debug/ios
+        bin_path.pop(); // Remove the current executable name: //Users/~/mopro/target/debug
+        bin_path.pop(); //Users/~/mopro/target/
+        bin_path.pop(); //Users/~/mopro
+        bin_path.push("mopro-ffi");
+        bin_path
+    };
```
vivianjeng, line 43: I am thinking to move the `uniffi-bindgen` to `test-e2e` and other apps that build bindings
(though `mopro-ffi` will not generate bindings anymore 👀)

or in this case
when I want to use `mopro` cli to generate templates for mobile apps
I have to 
1. install the whole `mopro` repo 
2. fix `uniffi-bindgen` path 
    ( I cannot easily use `mopro init` in other directories, only `test-e2e` works)

How do you think?

docs/docs/setup/rust-setup.md
```diff
@@ -61,25 +61,29 @@ Now you should copy your wasm and zkey files somewhere in the project folder. Fo
...
     // This file must be written to ./src
-    std::fs::write("./src/mopro.udl", mopro_ffi::app_config::UDL).expect("Failed to write UDL");
+    let udl_path = Path::new("src/mopro.udl");
+    if !udl_path.exists() {
+        std::fs::write(udl_path, mopro_ffi::app_config::UDL).expect("Failed to write UDL");
+    }
     // Finally initialize uniffi and build the scaffolding into the
     // rust binary
-    uniffi::generate_scaffolding("./src/mopro.udl").unwrap();
+    uniffi::generate_scaffolding(udl_path).unwrap();
```
vivianjeng, line 86: same here

cli/src/template/init/build.rs
alxkzmn: I think CLI needs some tests that would compile these files, should I create an issue for that?

mopro-ffi/src/app_config/android.rs
alxkzmn: I don't think I understand. Do you mean that this way of composing the relative path wouldn't work for a general case?

alxkzmn: I don't understand what you mean by "fixing the uniffi-bindgen path".

vivianjeng: >  Do you mean that this way of composing the relative path wouldn't work for a general case?

Yes
We will need to clone the `mopro` repo and use the `uniffi-bindgen` [here](https://github.com/zkmopro/mopro/blob/61f6b364bf9052a8d83251a7a4104b9ff541dede/mopro-ffi/Cargo.toml#L16)

I imagine the user can download the `mopro-cli` binary (e.g. like a circom [binary](https://github.com/iden3/circom/releases/tag/v2.2.1))
and execute `mopro init` in any directory
and with the `mopro init`, there will be a `uniffi-bindgen` in the new directory

cli/src/template/init/build.rs
vivianjeng: sounds good to me!
and anything you think we can integrate with `clippy` you can also create an issue. 🙏🏻

mopro-ffi/src/app_config/android.rs
alxkzmn: I think I understand the problem now. It would be solved if there was a way to call `uniffi::uniffi_bindgen_main()` directly with arguments.

mopro-ffi/src/app_config/ios.rs
```diff
@@ -88,12 +86,23 @@ pub fn build() {
         lib_out
     };
 
-    write_bindings_swift(&swift_bindings_dir);
+    generate_bindings(
+        (manifest_dir.clone() + "/src/mopro.udl").as_str().into(),
```
sifnoc, line 90: Does not need to `clone` in here.

sifnoc (changes requested):
I’d like to suggest a small change to help distinguish the original `mopro.udl` in `mopro-ffi`. Perhaps we could rename it to something like `mopro_example.udl` instead of `mopro.udl`?

alxkzmn:
@sifnoc  I think if we want to introduce `udl`s with different names, we would need to pass the name as a command line argument. This can be done as a separate issue if we need it.

mopro-ffi/src/app_config/ios.rs
alxkzmn: Done, thanks for noticing, it was a leftover :)

sifnoc:
> @sifnoc I think if we want to introduce `udl`s with different names, we would need to pass the name as a command line argument. This can be done as a separate issue if we need it.

Agree. There are at least three places where `mopro.udl` files are accessed: `mopro-ffi`, `test-e2e`, and the `mopro-example-app` generated by the mopro CLI.
However, it seems there is not enough documentation on how to modify and update the bindings. 
We can address this in another issue.

sifnoc (approved):

mopro-ffi/src/app_config/android.rs
```diff
@@ -2,9 +2,9 @@ use std::fs;
 use std::path::Path;
 use std::process::Command;
 
-use super::{cleanup_tmp_local, install_arch, install_ndk, mktemp_local};
+use uniffi::{generate_bindings, KotlinBindingGenerator};
```
chancehudson, line 5: After I wrote this rust code I started separating imports by line. It makes the file a bit more clean IMO and more resistant to conflicts. Totally optional

```suggestion
use uniffi::generate_bindings;
use uniffi::KotlinBindingGenerator;
```

mopro-ffi/src/app_config/android.rs
```diff
@@ -2,9 +2,9 @@ use std::fs;
...
 
-pub const MOPRO_KOTLIN: &str = include_str!("../../KotlinBindings/uniffi/mopro/mopro.kt");
+use super::{cleanup_tmp_local, install_arch, install_ndk, mktemp_local};
```
chancehudson, line 7:
```suggestion
use super::cleanup_tmp_local;
use super::install_arch;
use super::install_ndk;
use super::mktemp_local;
```

mopro-ffi/src/app_config/ios.rs
```diff
@@ -2,11 +2,9 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-use super::{cleanup_tmp_local, install_arch, mktemp_local};
+use uniffi::{generate_bindings, SwiftBindingGenerator};
```
chancehudson, line 5:
```suggestion
use uniffi::generate_bindings;
use uniffi::SwiftBindingGenerator;
```

mopro-ffi/src/app_config/ios.rs
```diff
@@ -2,11 +2,9 @@ use std::fs;
...
 
-pub const MOPRO_SWIFT: &str = include_str!("../../SwiftBindings/mopro.swift");
-pub const MOPRO_FFI_H: &str = include_str!("../../SwiftBindings/moproFFI.h");
-pub const MOPRO_MODULEMAP: &str = include_str!("../../SwiftBindings/moproFFI.modulemap");
+use super::{cleanup_tmp_local, install_arch, mktemp_local};
```
chancehudson, line 7:
```suggestion
use super::cleanup_tmp_local;
use super::install_arch;
use super::mktemp_local;
```

mopro-ffi/src/app_config/ios.rs
```diff
@@ -115,27 +124,33 @@ pub fn build() {
...
     // Copy the mopro.swift file to the output directory
-    cleanup_tmp_local(&build_dir_path)
+    cleanup_tmp_local(build_dir_path)
 }
 
-pub fn write_bindings_swift(out_dir: &Path) {
-    if let Ok(info) = fs::metadata(out_dir) {
-        if !info.is_dir() {
-            panic!("out_dir exists and is not a directory");
+fn rename_module_map_recursively(bindings_out: &PathBuf) {
```
chancehudson, line 145: Could you write a comment explaining what this function does? From context it moves a file named `moproFFI.modulemap` to `module.modulemap` in the same directory. It does this exactly once, so it doesn't fully traverse the directory structure if not needed.

mopro-ffi/src/app_config/ios.rs
```diff
@@ -115,27 +124,33 @@ pub fn build() {
...
+    for entry in fs::read_dir(bindings_out).expect("Failed to read bindings out directory") {
+        let entry = entry.expect("Failed to read entry");
+        let path = entry.path();
+        if path.is_file() && path.file_name().unwrap() == "moproFFI.modulemap" {
+            let dest_path = path.with_file_name("module.modulemap");
+            fs::rename(&path, &dest_path).expect("Failed to rename module map");
```
chancehudson, line 151: If we're only looking for 1 file an explicit `break` or `return` might be good here.

mopro-ffi/src/app_config/android.rs
```diff
@@ -32,10 +32,20 @@ pub fn build() {
...
+    generate_bindings(
+        (manifest_dir + "/src/mopro.udl").as_str().into(),
+        None,
+        KotlinBindingGenerator,
+        Some(bindings_out.to_str().unwrap().into()),
+        None,
+        None,
+        false,
+    )
+    .expect("Failed to generate bindings");
```
chancehudson, line 47: This is not something we should do in this PR, but in the future we should probably refactor to use `anyhow::Result` as the return type in all these functions (including `main`). In all the rust codebases i've worked in (which to be fair is only like 4) I've had to refactor to add `anyhow::Result` everywhere.

If there's an error case we want to handle explicitly we can always do something like

```rs
if let Err(error) = fallible_operation() {
    println!("Operation did not succeed because of reason: {}", error);
    std::process::exit(1);
}
```

chancehudson (approved):

chancehudson:
Looks good to me, good PR description too. I left comments about code style only.

mopro-ffi/src/app_config/android.rs
alxkzmn: I'd love to automate this, is there a linter for that?

mopro-ffi/src/app_config/ios.rs
alxkzmn: We're looking for multiple as we're normally building for both physical device and the simulator (each corresponding folder has a `modulemap`). I'll add the comment.

PR #267, opened 2024-12-05
`mopro create` improvements

alxkzmn:
The scope of the PR is to add the functionality to create Flutter and React Native apps using mopro CLI.
Changes:
- `mopro create` now contains the options for Flutter and React Native templates;
- improved error handling in some cases (e.g., more explicit message when a directory exists and cannot be overwritten).

Issue #250, opened 2024-10-31
Customize UDL file


vivianjeng:
# Problem
Currently we fix the uniffi files and functions in [`bin/ios.rs`](https://github.com/zkmopro/mopro/blob/main/mopro-ffi/src/app_config/ios.rs) and [`bin/android.rs`](https://github.com/zkmopro/mopro/blob/main/mopro-ffi/src/app_config/android.rs)
And if we create a new app like [`test-e2e`](https://github.com/zkmopro/mopro/tree/main/test-e2e)
We cannot update the functions through `mopro.udl` because it will be overwritten by `bin/ios`/`bin/android`
It will be difficult for applications that need to extend the rust package
e.g. GPU research or if we will build other SDKs

### mopro-ffi
- remove fixed bindings: e.g. [mopro.swift](https://github.com/zkmopro/mopro/blob/dc5609bb15606cdbbb1cb80f2f587b32a9908dc0/mopro-ffi/src/app_config/ios.rs#L7)
- generate bindings in the script (`bin/ios`, `bin/android`): e.g. [build_bindings.sh](https://github.com/zkmopro/mopro/blob/dc5609bb15606cdbbb1cb80f2f587b32a9908dc0/mopro-ffi/build_bindings.sh#L8)
- copy the bindings to the output folder (e.g.`MoproiOSBindings`)

### test-e2e
- users can overwrite the `src/mopro.udl` file and output customized bindings with `cargo run --bin ios` and `cargo run --bin android`

## Acceptance criteria
- The new bindings can be added in [`test-e2e/src/lib.rs`](https://github.com/zkmopro/mopro/blob/main/test-e2e/src/lib.rs)
- both mopro-ffi bindings (e.g. `generateCircomProof`) and new bindings can be called in both iOS and Android 

## Next steps (optional)
- publish the new crate
- update mopro cli

chancehudson:
I think we discussed this a bit before, but we should consider if we should recommend teams move to their own FFI implementation. Sorta like ejecting from an expo app, but ejecting from mopro.

For most simple cases they would use mopro, but if they want to customize the UDL (by adding new functions/types) it would be better for project to eject from mopro so they can edit things without abstraction.

vivianjeng:
@alxkzmn I imagine that
developers can use both functions in `mopro-ffi` e.g. `generateCircomProof`
and also functions defined in like `test-e2e` e.g. `YourStruct`
[...]

PR #268, opened 2024-12-09
Mopro cli supporting web platform

sifnoc:
There is  issue to use wasm-pack build script similar like https://github.com/rustwasm/wasm-pack/issues/1400 

There is an issue with using the `wasm-pack` build script, similar to rustwasm/wasm-pack#1400.
- Created two build scripts:
  - **cli/src/template/init/src/bin/web.rs** 
  - **test-e2e/src/bin/web.rs**  
  These scripts use the `wasm-pack` CLI for compiling WASM with `std::process`.
- Added the "mopro-wasm" module to cli/src/template due to the bug.

I left "test-e2e/src/bin/web.rs" for consistency with other platforms (e.g., iOS, Android), but note that using a simple js script (yarn build) in "test-e2e/web/package.json" would be significantly faster.

Since only halo2 wasm is supported in "mopro-wasm", a notification has been added for cases where the "web" platform is selected in the cli. In the future, these selectable combinations should be handled in a more generalized way.

This PR will be convert to review after merge https://github.com/zkmopro/mopro/pull/267


cli/src/print.rs
```diff
@@ -61,6 +61,18 @@ pub(crate) fn print_create_android_success_message() {
...
+pub(crate) fn print_create_web_success_message() {
+    print_green_bold("Template created successfully!".to_string());
+    println!();
+    print_green_bold("Next steps:".to_string());
+    println!();
+    print_green_bold("  You can now use the following command to start web server:".to_string());
+    println!();
+    print_bold(r"    cd web && yarn start".to_string());
+    println!();
+    print_green_bold("This will start a simple web server for your browser.".to_string());
```
vivianjeng, line 73: now we have the page for flutter and react native like
https://github.com/zkmopro/mopro/blob/a95b7f7ee81e8a111b321f4ef763d4d129cede65/cli/src/print.rs#L80

I think we can also maintain this setup page for website

test-e2e/src/bin/web.rs
```diff
@@ -1,4 +1,64 @@
-// TODO: copying pkg that includes wasm files from mopro-wasm
+use std::{path::Path, process::Command};
+
 fn main() {
-    println!("Perform wasm files from mopro-wasm");
```
vivianjeng, line 3: I am thinking about this
can it be like binaries in iOS and android
e.g.
```rs
mopro_ffi::app_config::ios::build();
```

and we can maintain a crate called `mopro_wasm`
and it just call the 
```rs
mopro_wasm::app_config::web::build();
```

so the e2e apps doesn't have to maintain the logic of `mopro_wasm` build (when we upgrade the script)
so we need a file in `mopro-wasm/src/app_config` like in `mopro-ffi/src/app_config`
to maintain the logic

cli/src/template/init/mopro-wasm/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "mopro-wasm"
```
vivianjeng, line 2: Is it possible that we can just reuse the `mopro-wasm` crate in mopro
e.g.
```toml
[dependencies]
mopro-wasm = { git = "https://github.com/zkmopro/mopro" }
```
(maybe we also add a feature flag like `mopro-ffi` to have a `halo2` feature flag)

Ideally we want to use the same crate in the root
(can it select different toolchain for different targets?
I think it would be also fine if we use nightly toolchain for `mopro-ffi` buildings?)

vivianjeng (commented):
I think ideally the structure should be like `test-e2e`
like
```sh
test-e2e
├── Cargo.toml
├── MoproAndroidBindings
├── MoproiOSBindings
├── MoproWASMBindings
├── SwiftBindings
├── android
├── build.rs
├── ios
├── src
│   ├── bin
│   ├── lib.rs
│   └── mopro.udl
├── tests
└── web
```
and hopefully it can work like https://github.com/zkmopro/mopro/issues/250
let users customize `src/lib.rs` to export customized `wasm_bindgen` functions

and use `mopro build` for `web`
it can generate another new `MoproWASMBindings`

I am not sure if it is possible to fix or how long does it take to fix
but actually the current structure works for me
if you think it will take much effort to fix this
you can create another issue and PR like #250

cli/src/print.rs
sifnoc: Okay, I will add the setup page name as "Web(WASM) Setup" and also add `Prerequisites` for installing `wasm-pack`.

cli/src/template/init/mopro-wasm/Cargo.toml
sifnoc: That's exactly what I want to achieve, but I haven't been able to make it work. I need to understand more about how `wasm-pack` works before attempting to write a lower-level build script.
In the meantime, it might be worth trying your suggestion of using the nightly toolchain for mopro-ffi and the CLI.

sifnoc:
I was exploring the possibility of using a build script instead of externally invoking `wasm-pack` with `Command`. I concluded that while it’s not impossible, there’s no clean solution, and using a build script feels less hacky.

Notably, my exploration ended with this discussion: https://github.com/rustwasm/wasm-bindgen/discussions/3494. It addresses the pre- and post-compilation processes in `wasm-bindgen`. while workarounds exist, there is still no clean, native cargo solution for this.

As a result, I decided to stick with externally invoking `wasm-pack`. additionally, I updated the web build script to fetch `mopro-wasm` using `git` instead of including `mopro-wasm` directly in the template (it has now been removed from the template).

workflow for compiling wasm for web in the cli looks like this:
1. **"cli/build"** calls the `build` command in **"mopro-wasm/app-config/web"**.  
2. The **"web"** fetches "mopro-wasm" using 'git'.  
3. The **"web"** compiles "mopro-wasm" in the fetched directory.

vivianjeng:
I found a problem
If I create a project without building for iOS and Android
It will not copy `mopro.udl` file
but the `build.rs` still compiles for it
```sh
mopro-example-app mopro build
✔ Build mode · release
✔ Select platform(s) to build for (multiple selection with space) · web
? Halo2 WASM code will only be generated for the web platform. Do you want
✔ Halo2 WASM code will only be generated for the web platform. Do you want to continue? · yes
thread 'main' panicked at cli/src/build.rs:149:22:
internal error: entered unreachable code
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

mopro-wasm/src/app_config/web.rs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        // When running the script from `test-e2e`
+        cwd.parent()
+            .expect("Failed to get parent directory")
+            .join("mopro-wasm")
+    } else {
+        // When running the script from the CLI template or else, download 'mopro-wasm' from repo
+        fetch_mopro_wasm();
+        cwd.join("mopro-wasm")
+    };
```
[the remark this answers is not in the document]
vivianjeng: In this case I would suggest that you create a PR for `mopro-wasm/src/app_config/web.rs` first
and then complete the CLI using
```toml
mopro-wasm = { git = "https://github.com/zkmopro/mopro" }
```
so that you don't even have to clone the mopro repo to use `mopro-wasm`

vivianjeng (commented):
I found that instead of cloning the whole `mopro-wasm`
you can create an *environment* to use `mopro-wasm`
e.g. you setup the settings like mopro-wasm
with `.cargo/`, `rust-toolchain.toml` and import crates like
```toml
[package]
name = "mopro-wasm-lib"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["rlib", "cdylib"]

[dependencies]
mopro-wasm = { path = "../../mopro/mopro-wasm", features = ["gemini", "hyperplonk", "plonk"]  }
# can be switched to git path

[target.wasm32-unknown-unknown.dependencies]
# ...

```

and you just import the `mopro-wasm` functions like

```rs
// lib.rs
use mopro_wasm::halo2::gemini;
use mopro_wasm::halo2::hyperplonk;
use mopro_wasm::halo2::plonk;
```

It will create the wasm for `generate_proof` and `verify_proof`

(You should also make the module public)

so you don't have to switch to the path of `mopro-wasm` and use `wasm-pack` to generate bindings

vivianjeng:
I created a PR for the suggestions https://github.com/zkmopro/mopro/pull/276
Please check 🙏🏻

mopro-wasm/src/app_config/web.rs
```diff
@@ -0,0 +1,114 @@
...
+        let package_name = "mopro-wasm";
+
+        // Fetch without checkout for avoiding download all
+        let status = Command::new("git")
+            .args(["clone", "--no-checkout", repo_url])
+            .status()
+            .expect("Failed to execute git clone");
+        if !status.success() {
+            panic!("Failed to clone the repository.");
+        }
```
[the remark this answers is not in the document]
sifnoc: I like this approach! will update it.

test-e2e/Cargo.toml
```diff
@@ -13,8 +13,12 @@ name = "ios"
 [[bin]]
 name = "android"
 
+[[bin]]
+name = "web"
+
 [dependencies]
 mopro-ffi = { path = "../mopro-ffi", features = ["halo2", "circom", "ashlang"] }
+mopro-wasm-lib = { path = "./mopro-wasm-lib" }
```
vivianjeng, line 21: here should be `mopro-wasm = { path = "../mopro-wasm" }`
because you need to use `mopro_wasm::app_config::web::build()`
(it should work just like `mopro-ffi`)

`mopro-wasm-lib` is used for wasm library / mostly for the nightly environment

test-e2e/mopro-wasm-lib/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub use mopro_wasm::app_config;
```
vivianjeng, line 1: so you don't have to export here

test-e2e/web/package.json
```diff
@@ -6,6 +6,7 @@
   "author": "Jinhwan",
   "license": "MIT",
   "scripts": {
+    "build": "cd ../../mopro-wasm && wasm-pack build --target web --out-dir ../test-e2e/web/mopro-pkg -- --all-features",
```
vivianjeng, line 9: you can try to use `cargo run --bin web` in the build script

cli/src/template/init/Cargo.toml
```diff
@@ -14,6 +14,7 @@ name = "mopro_bindings"
 default = ["<FEATURES>"]
 
 [dependencies]
+mopro-wasm = { git = "https://github.com/zkmopro/mopro", branch = "mopro-cli-web" }
```
vivianjeng, line 17: TODO: fix this branch

sifnoc: I will fix this in another new PR.

vivianjeng (approved):
LGTM
NOTE: there will be another PR fix the cli
https://github.com/sifnoc/mopro/tree/mopro-cli-web-fix

PR #270, opened 2024-12-12
CLI can select target arch

sifnoc:
- `mopro-ffi:app_config::ios::build` and `mopro-ffi:app_config::android::build` now require `target_arch` as an input instead of being hardcoded value inside the function.
- The build scripts in `cli/src/template/init/bin/*` and `test-e2e/src/bin/*` read the `target_arch` value from the environment variable. If no value is provided, all available architectures are selected by default.
- Did not add `mopro-ffi` crate in the cli for only reading `*_ARCHS` variables. It's too heavy to compile for simple.

### DIFF
--- a/cli/src/template/init/src/bin/ios.rs
+++ b/cli/src/template/init/src/bin/ios.rs
@@ -1,21 +1,20 @@
+use mopro_ffi::app_config::ios::IOS_ARCHS;
+
 fn main() {
-    let available_archs = vec!["aarch64-apple-ios", "aarch64-apple-ios-sim", "x86_64-apple-ios"];
-    
     let ios_archs: Vec<String> = if let Ok(ios_archs) = std::env::var("IOS_ARCHS") {
         ios_archs.split(',').map(|arch| arch.to_string()).collect()
     } else {
         // Default case: select all supported architectures if none are provided
-        available_archs.iter().map(|&arch| arch.to_string()).collect()
+        IOS_ARCHS.iter().map(|&arch| arch.to_string()).collect()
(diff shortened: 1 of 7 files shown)


cli/src/template/init/src/bin/ios.rs
```diff
@@ -1,5 +1,23 @@
...
+            IOS_ARCHS.contains(&arch.as_str()),
+            "Unsupported architecture: {}",
+            arch
+        );
+    }
+
     // A simple wrapper around a build command provided by mopro.
     // In the future this will likely be published in the mopro crate itself.
-    mopro_ffi::app_config::ios::build();
+    mopro_ffi::app_config::ios::build(&ios_archs);
```
vivianjeng, line 22: I think passing `archs`  as inputs doesn't make much sense to me
and I think letting users maintaining the line 4-18 is too much work
(and we have to maintain double files in template and test-e2e, and also documentation)

if it is in `env`, it can just let `mopro_ffi::app_config::ios::build()` to read the `env` and handle the error assertions?
like the debug/release mode, it is also handled by `mopro_ffi::app_config::ios::build()`

vivianjeng (commented):
I think ideally only `cli/src/build.rs`, `mopro-ffi/src/app_config/*` needs to be changed.
let me know if it is possible to do this 🙏🏻

sifnoc:
This PR will be replace another new PR with https://github.com/zkmopro/mopro/tree/selectable_target_arch

sifnoc:
Replaced #271

PR #276, opened 2024-12-18
suggestions

vivianjeng:
- It creates an environment for `wasm-pack` called `mopro-wasm-lib`
  - so the `cargo run --bin web` is executing `wasm-pack` within `mopro-wasm-lib`
  - and it imports the halo2 functions like `use mopro_wasm::halo2::plonk;` if the feature is activated
  - (also write down an example `hello()` to make it flexible, scalable) 
 
> To avoid the same name as `mopro-wasm`, the crate is called `mopro-wasm-lib`

- **TODO:** 
  - update cli templates


sifnoc:
It's applied https://github.com/zkmopro/mopro/pull/268/commits/a6ee961e277b871eebd27a4d0886288365476bcc and merged at #268.

PR #277, opened 2024-12-19
Fixed mopro cli for web platform

sifnoc:
- Added "mopro-wasm-lib" in template 
- "mopro-wasm-lib" is use "mopro-wasm" in main branch

### DIFF
--- a/cli/src/build.rs
+++ b/cli/src/build.rs
@@ -146,16 +146,11 @@ pub fn build_project(
 
             let cwd = std::env::current_dir().unwrap();
             let target_dir = &cwd.join("mopro-wasm-lib");
-            if target_dir.exists() {
-                style::print_yellow(
-                    "'mopro-wasm-lib' already exist, Please remove it and try again".to_string(),
-                );
-                return Err(anyhow::anyhow!(""));
+            if !target_dir.exists() {
+                const WASM_TEMPLATE_DIR: Dir =
+                    include_dir!("$CARGO_MANIFEST_DIR/src/template/mopro-wasm-lib");
+                copy_embedded_dir(&WASM_TEMPLATE_DIR, &target_dir)?;
             }
-
-            const WASM_TEMPLATE_DIR: Dir =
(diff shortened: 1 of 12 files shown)


cli/src/build.rs
```diff
@@ -136,14 +139,19 @@ pub fn build_project(
                 .default(true)
                 .interact()?;
 
+            const WASM_TEMPLATE_DIR: Dir =
+                include_dir!("$CARGO_MANIFEST_DIR/src/template/mopro-wasm-lib");
+            let cwd = std::env::current_dir().unwrap();
+            copy_embedded_dir(&WASM_TEMPLATE_DIR, &cwd.join("mopro-wasm-lib"))?;
```
vivianjeng, line 145: actually it should not overwrite the `mopro-wasm-lib`
it should check if it exists, and not copy again if it exists

vivianjeng: we want users can write their customized functions to use bindgen in this folder
e.g.
```rs
#[wasm_bindgen]
pub fn hello() -> Result<JsValue, JsValue> {
    Ok(JsValue::from_str("Hello, world!"))
}
```
so it should not be overwritten

vivianjeng (changes requested):

vivianjeng:
Here should be updated as well
https://github.com/zkmopro/mopro/blob/9cbce440321ce44501fc843a7072abe7cec0f01b/cli/src/template/init/Cargo.toml#L17

vivianjeng (changes requested):
please check 🙏🏻
Thank you

test-e2e/mopro-wasm-lib/Cargo.toml
```diff
@@ -7,11 +7,11 @@ edition = "2021"
 crate-type = ["rlib", "cdylib"]
 
 [dependencies]
-mopro-wasm = { path = "../../mopro-wasm", features = [
+mopro-wasm = { git = "https://github.com/zkmopro/mopro",features = [
```
vivianjeng, line 10:
```suggestion
mopro-wasm = { path = "../../mopro-wasm", features = [
```

Issue #264, opened 2024-12-03
Mopro CLI configuration file


vivianjeng:
# Problem
Now when users use mopro cli to 
`init`: choose proving systems
`build`: choose target platforms
`create`: create templates -> but the CLI doesn't know which proving system and which target platform the users choose before

## Details
We need a file like `Config.toml` in the root dir (when executing `mopro init`)
It stores the config that the user chooses before e.g.
```sh
proving_system = ["circom", "halo2"]
target = ["ios", "android"]
```
then the `mopro create` can read this config file and create customized templates for it

## Acceptance criteria
A config file stores when executing `mopro init`
read this config file and create customized templates when executing `mopro create`

PR #279, opened 2024-12-19
chore: improve mopro-wasm

vivianjeng:
- halo2 crate should be optional
- It will show the building logs immediately


sifnoc (approved):
I added small addition for ignore warn.
I also realized that `plonk-fibonacci` is added to the cargo file in the mopro-example-app whenever "halo2" is selected, without considering the platform selection. I will address this in #264 

your improvement looks good to me. you can revert my addition if you don't mind.

PR #282, opened 2024-12-20
Added config file for CLI

sifnoc:
Related issue #264 

- Create `Config.toml`  when `mopro init`
- Use it `mopro build` and `mopro create`

The template selection is more informative like one below: 
```
✔ Create template · ios          - Require bindings
Cannot create ios template - build binding first
? Create template ›
  ios          - Require bindings
  android
  web          - Require bindings
  flutter      - Requires ios bindings
  react-native - Requires ios bindings
  ```

### CODE REVIEW

File: cli/src/build.rs [hunk 1]
```diff
@@ -183,7 +201,13 @@ pub fn build_project(
             }
         }
 
-        print_binding_message(platforms)?;
+        // Write selected platforms for create command
+        for p in &selected_platforms {
+            config.target_platforms.push(p.to_string())
+        }
+        write_config(&config_path, &config)?;
```
vivianjeng, line 208: when I execute `mopro build` for iOS and `mopro create` for iOS
and execute again `mopro build` for Android and `mopro create`
The `target_platforms` is overwritten
it should be `target_platforms = ["ios","android"]` but it shows `target_platforms = ["android"]`
(so I cannot create for react-native even if I have iOS bindings)

try to read the config first and push the config if needed?

cli/src/create.rs
```diff
@@ -319,7 +374,7 @@ fn copy_dir(input_dir: &Path, output_dir: &Path) -> anyhow::Result<()> {
     Ok(())
 }
 
-fn copy_keys(target_dir: std::path::PathBuf) -> Result<(), anyhow::Error> {
+fn copy_keys(target_dir: std::path::PathBuf) -> anyhow::Result<(), Error> {
```
vivianjeng, line 377: I think it can be
```suggestion
fn copy_keys(target_dir: std::path::PathBuf) -> anyhow::Result<()> {
```
?

sifnoc:
I assumed that users will not manually remove the *Bindings folder in the initialized project created with `mopro init`.
Therefore, once a `selected platform` is written to the config file, it will not be removed.

cli/src/create.rs
```diff
@@ -85,10 +87,63 @@ pub fn create_project(arg_platform: &Option<String>) -> anyhow::Result<()> {
...
+            }
+        }
+
+        if config.target_platforms.contains(template) {
+            items.push(template.to_string());
+            unselectable.push(false);
+        } else {
+            items.push(format!("{:<12} - Require binding", template));
+            unselectable.push(true);
+        }
```
vivianjeng, line 145: I found the correct logic is like this
You should display the result of `react-native` and `flutter` if it is selectable
```suggestion
            // continue;
            } else {
                items.push(template.to_string());
                unselectable.push(false);
            }
        } else if config.target_platforms.contains(template) {
            items.push(template.to_string());
            unselectable.push(false);
        } else {
            items.push(format!("{:<12} - Require binding", template));
            unselectable.push(true);
        }
```

sifnoc:
> Kimi changes the term `platform` to `framework` so I think maybe you need to merge the `main` branch to see the latest changes 🙏🏻

I believe the term 'framework' doesn't fit well in the 'build' phase. We've been using the term 'template' in the 'create' phase, so I decided to keep 'template' for printout strings in the 'create' phase while using 'framework' in the code for consistency.

vivianjeng (approved):
LGTM 
thank you!

PR #283, opened 2024-12-26
doc: update android/iOS benchmark on readme

### LINKED ISSUE #44: Benchmark and compare with rapidsnark prover
## Problem

Right now we are using ark-groth16 as native prover. This is in Rust and simplifies a lot of things. At the same time, rapidsnark is more mature and with recent optimizations for mobile it might be more performant.

We want to understand the performance diff better, understand where it makes sense and possibly have it as a feature that can be used (switching between provers to use).

## Details

See https://github.com/iden3/rapidsnark

Note that currently witness calculation is bottleneck (wasm), not prover. This means this issue isn't super high on the priority list right now. It may become more relevant in the future, especially as prover time is a bottleneck and if it has a bi impact on e.g. older Android devices.

Polygon Identity might also do some benchmarking of Keccak256 using their stack so we get something roughly comparable (albeit perhaps with different devices).

## Acceptance criteria

Better data on performance diff for using rapidsnark vs ark-groth16.

### LINKED ISSUE #266: Benchmark for Android
# Problem
Now we have benchmark results for iOS and Mac here: https://zkmopro.org/docs/performance/
and here https://github.com/zkmopro/mopro?tab=readme-ov-file#performance
but Android benchmark results are missing

## Details

We can have a simplified result table like this
<table>
  <tr>
    <th>Circuits</th>
    <th>snarkjs groth16 fullprove</th>
    <th>mopro (rust-witness + ark-works)</th>
    <th>witnesscalc + rapidsnark</th>
  </tr>
  <tr>
    <td>Keccak256</td>
    <td>- ms</td>
    <td>- ms</td>
    <td>- ms</td>
  </tr>
  <tr>
    <td>SHA256</td>
    <td>- ms</td>
    <td>- ms</td>
    <td>- ms</td>
  </tr>
  <tr>
    <td>RSA</td>
    <td>- ms</td>
    <td>- ms</td>
    <td>- ms</td>
  </tr>
  <tr>
    <td>Semaphore</td>
    <td>- ms</td>
    <td>- ms</td>
    <td>- ms</td>
  </tr>
  <tr>
    <td>Anon Aadhaar</td>
    <td>- ms</td>
    <td>- ms</td>
    <td>- ms</td>
  </tr>
</table>

If there is a result for separating witness generation + proof generation it would be great

## Acceptance criteria
Update benchmark result in README and performance page
(if there is an app like https://github.com/zkmopro/benchmark-app for Android would be great)

## Next steps (optional)
Describe the subsequent actions to be taken.

vivianjeng:
close #266 
close #44 
ref: https://github.com/zkmopro/android-benchmark-app

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,40 +45,79 @@ Check the details here: [performance](https://zkmopro.org/docs/performance).
 
 <table>
   <tr>
-    <th>Circuits</th>
+    <th>Android (Samsung S23U)</th>
     <th>snarkjs groth16 fullprove</th>
     <th>mopro (rust-witness + ark-works)</th>
     <th>witnesscalc + rapidsnark</th>
   </tr>
   <tr>
     <td>Keccak256</td>
-    <td>8406.2 ms</td>
-    <td>1381.8 ms (~6x)</td>
-    <td>2792.6 ms (~3x)</td>
+    <td>6.584 s</td>
+    <td>2.55 s (~2.5x)</td>
+    <td>0.867 s (~7.6x)</td>
   </tr>
   <tr>
     <td>SHA256</td>
-    <td>2537.6 ms</td>
-    <td>640.7 ms (~4x)</td>
-    <td>817.5 ms (~3.1x)</td>
+    <td>2.499 s</td>
+    <td>1.73 s (~1.4x)</td>
+    <td>0.252 s (~9.9x)</td>
   </tr>
   <tr>
     <td>RSA</td>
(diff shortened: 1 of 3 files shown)


vivianjeng:
https://docs.google.com/spreadsheets/d/1TDgL2NXxYl8UH-RZPWfWdawY0tjxf3c6l8B11_FG7Kg/edit?usp=sharing

PR #287, opened 2025-01-06
Refactor/replace plain str with enum

KimiWu123:
### Two major changes in this PR
1. Using `enum` to replace `String` (or `&str`) in `create.rs`
2. Refactor `create.rs`, moved different platforms to corresponding files (under `create/` folder)
----

### Description
My intention in the first place is to replace `String` with `enum` only. However, during this refactor, I realized there are more things I can do and that's why I refactored `create.rs`. These two changes separated in different commits. If anyone feels any one of them is not okay, then it's easy to remove it.
Most of changes were just moving from one place to another without (or with limited) changes.

### DIsccusion
- Naming: I renamed `platform` to `APP`, the definition of original `platform` looks more like apps or languages for me. That's why I rename the variable. `framework` might be also a good candidate too. Any feedback is welcome.

### DIFF
--- a/cli/src/create/utils.rs
+++ b/cli/src/create/utils.rs
@@ -13,7 +13,7 @@ use indicatif::ProgressStyle;
 use reqwest::blocking::Client;
 use zip::ZipArchive;
 
-use super::APP;
+use super::Framework;
 
 pub fn copy_android_bindings(
     android_bindings_dir: &Path,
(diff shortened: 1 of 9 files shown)


cli/src/create/utils.rs
```diff
@@ -0,0 +1,205 @@
...
+use anyhow::{Error, Result};
+use include_dir::include_dir;
+use include_dir::Dir;
+use indicatif::ProgressBar;
+use indicatif::ProgressStyle;
+use reqwest::blocking::Client;
+use zip::ZipArchive;
+
+use super::APP;
+
```
KimiWu123, line 17: All the functions here were copied from `create.rs` without any changes except `check_bindings`

cli/src/print.rs
```diff
@@ -34,81 +34,3 @@ pub(crate) fn print_build_success_message() {
     print_bold("   mopro create".to_string());
     print_footer_message();
 }
-
-pub(crate) fn print_create_ios_success_message() {
```
KimiWu123, line 38: All the message printing functions were moved to corresponding files.

KimiWu123:
> It works well for me! Thank you And I prefer the name `framework` instead of `APP` I think the `App` sounds too general so it is confusing to me

fixed in [8e5cfa7](https://github.com/zkmopro/mopro/pull/287/commits/8e5cfa7ce4151d673d6760be79b5795c1a4b0a2d)

> 
> And please check the clippy with the command
> 
> ```shell
> cargo clippy --all-targets --all-features
> ```

fixed in [f5b627f](https://github.com/zkmopro/mopro/pull/287/commits/f5b627f795ec72f0aa51ae3a5d17fa1bf3284106)

vivianjeng (approved):

PR #288, opened 2025-01-06
doc: 2024 retro

vivianjeng:


docs/blog/2025-01-06-2024-retro.md
```diff
@@ -0,0 +1,118 @@
...
+or in kotlin
+
+```kotlin
+val hash = poseidon(input)
+```
+
+Additionally, this approach is compatible with WASM for browsers. You can define a function in Rust as follows:
+
+```rust
+use serde_wasm_bindgen::to_value;
```
sifnoc, line 83: It should be added this too
```
use wasm_bindgen::prelude::JsValue;
```

docs/docusaurus.config.ts
```diff
@@ -27,6 +27,13 @@ const config: Config = {
           includeCurrentVersion: true,
           lastVersion: '0.1.0',
         },
+        blog: {
+          showReadingTime: true,
+          // Please change this to your repo.
```
sifnoc, line 32: Is it left-over?

sifnoc (changes requested):

docs/docusaurus.config.ts
vivianjeng: [6755847](https://github.com/zkmopro/mopro/pull/288/commits/675584747e16b857c3f8e6d710c339e2faa7005e)

docs/blog/2025-01-06-2024-retro.md
vivianjeng: [6755847](https://github.com/zkmopro/mopro/pull/288/commits/675584747e16b857c3f8e6d710c339e2faa7005e)

PR #297, opened 2025-01-14
Refactor/combine strings in CLI and mopro-ffi

### LINKED ISSUE #296: Refactor plain strings in CLI and mopro-ffi by replacing the with `enum`
# Problem
We have many duplicated plain string in CLI.

## Details
Plain strings are difficult to maintain due to issues like typos or incorrect copy-pasting. Therefore, I propose replacing these strings with enum, consolidating the strings in a single location, and using the newly created enums throughout the codebase. This issue arises not only in the CLI module but also in mopro-ffi. Both modules use the same strings, but they are currently declared separately within each module.
To address this issue, my goals are:
1. Replace plain strings with enum in the codebase.
2. Unify string declarations for the CLI and mopro-ffi modules.

I plan to split this task into three PRs.
1. The first try to replace strings in CLI/create.rs: #287 
2. Refactor CLI #295 
3. Unify strings declarations for `CLI` and `mopro-ffi` modules, #297 

## Acceptance criteria
Everything should work as before.

KimiWu123:
close #296 , the last part of #296.


cli/src/constants.rs
```diff
@@ -1,113 +1,224 @@
...
     Aarch64Apple,
     Aarch64AppleSim,
     X8664Apple,
 }
 
-impl From<usize> for IosArch {
-    fn from(idx: usize) -> Self {
-        IOS_ARCHS[idx].into()
-    }
+pub struct IosArchInfo {
```
KimiWu123, line 59: Introduced `xxxInfo` to ensure the order of `enum` and `string` is consistent.

mopro-ffi/src/app_config/android.rs
```diff
@@ -2,61 +2,52 @@ use std::fs;
...
     } else {
         // Default case: select all supported architectures if none are provided
-        ANDROID_ARCHS.iter().map(|arch| arch.to_string()).collect()
+        AndroidArch::all_strings()
+            .iter()
+            .map(|s| AndroidArch::parse_from_str(s))
+            .collect()
     };
 
-    // Check 'ANDRIOD_ARCH' input validation
```
KimiWu123, line 48: removed because `AndroidArch::parse_from_str` will do the same check

cli/src/constants.rs
```diff
@@ -1,113 +1,224 @@
...
+            .iter()
+            .find(|info| info.mode == *self)
+            .map(|info| info.str)
+            .expect("Unsupported Mode, only support 'release' and 'debug'")
+    }
 
+    pub fn parse_from_str(s: &str) -> Self {
+        MODES
+            .iter()
+            .find(|info| info.str.to_lowercase() == s.to_lowercase())
```
KimiWu123, line 35: Want to raise a discussion about string comparison. In normal cases, we usuall convert strings to lower/upper case and then compare them. I'm not sure if that's the case here. All the `parse_from_str` implemeted in this file, I did the same thing.

mopro-ffi/src/app_config/constants.rs
```diff
@@ -0,0 +1,171 @@
...
+pub const ARCH_I686: &str = "x86";
+pub const ARCH_ARM_V7_ABI: &str = "armeabi-v7a";
+pub const ARCH_ARM_64_V8: &str = "arm64-v8a";
+
+pub const ENV_CONFIG: &str = "CONFIGURATION";
+pub const ENV_ANDROID_ARCHS: &str = "ANDROID_ARCHS";
+pub const ENV_IOS_ARCHS: &str = "IOS_ARCHS";
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Mode {
```
KimiWu123, line 12: moved `Mode` and `Arch` here due to module dependency. `mopro-ffi` should not depend on `CLI`.

cli/src/constants.rs
```diff
@@ -1,113 +1,56 @@
...
     Android,
     Web,
 }
 
+struct PlatformInfo {
+    platform: Platform,
+    str: &'static str,
+}
+
+const PLATFORMS: [PlatformInfo; 3] = [
```
KimiWu123, line 16: Introduced xxxInfo to ensure the order of enum and string is consistent.

cli/Cargo.toml
```diff
@@ -19,3 +19,4 @@ zip = "2.2.1"
 reqwest = { version= "0.12.9", features = ["blocking"] }
 indicatif = "0.17.9"
 serde = { version = "1.0.216", features = ["derive"] }
+mopro-ffi = {path = "../mopro-ffi"}
```
vivianjeng, line 22: I am still concerned about importing `mopro-ffi`
if we will just use the `constants` (which are pretty simple definitions)

I benchmark the difference between the two
without importing `mopro-ffi` is **9s** to install the CLI
with importing `mopro-ffi` is **36s** to install the CLI

I actually prefer we keep `mopro-ffi` and `cli` independent
and define the `constants` again in `cli`
how do you think? @sifnoc @KimiWu123

sifnoc: > I benchmark the difference between the two without importing `mopro-ffi` is **9s** to install the CLI with importing `mopro-ffi` is **36s** to install the CLI

That's huge

> I actually prefer we keep `mopro-ffi` and `cli` independent and define the `constants` again in `cli` how do you think? @sifnoc @KimiWu123

I think it would be fine to add a test case in the github workflow to verify that the two constant definitions are identical.
Would creating a new crate solely for constants be excessive? Both options seem fine to me.

KimiWu123: I like the idea to create a independent crate, but it's just simple definitions as Vivi said. I agree that we should make things simple. Will add another file in `CLI`.

mopro-ffi/src/app_config/android.rs
```diff
@@ -74,45 +64,46 @@ pub fn build() {
...
         "{}/{}/{}/libmopro_bindings.so",
         build_dir.display(),
-        arch,
-        mode
+        arch_str,
+        mode.as_str()
     ));
     let out_lib_dest = bindings_out.join(format!("jniLibs/{}/libuniffi_mopro.so", folder));
 
+    println!("COPY: {:?} -> {:?}", out_lib_path, out_lib_dest);
```
sifnoc, line 104: Just curious, is it left-over?

sifnoc (approved):
It looks fine to me for now. If anyone encounters issues syncing two `constants.rs` files in the future, we can consider to seperate the architecture definitions in `constants` of the cli and syncing them with the "constants.rs" file in `mopro-ffi` using a build script in the cli.

mopro-ffi/src/app_config/android.rs
KimiWu123: Nice catch, fixed in [9dd8197](https://github.com/zkmopro/mopro/pull/297/commits/9dd81972de930db585b601cac59aded0ac17f880).

vivianjeng (approved):
LGTM 
Thank you!

Issue #299, opened 2025-01-16
`circom-prover` spec


vivianjeng:
# Problem
Related to #224 
1. Current mopro depends on `ark-circom` version but we need to patch it to get the correct version: [zkmopro/circom-compat](https://github.com/zkmopro/circom-compat)
    - we replace `wasmer` with `rust-witness`. So we remove whole `wasmer` dependencies in zkmopro/circom-compat
    - There are a lot of problems with `wasmer` in mobile. Checkout [issues](https://github.com/zkmopro/mopro/issues?q=is%3Aissue+wasmer)
    - See: [Circom Comparison](https://zkmopro.org/blog/circom-comparison) for more details
2. We only use a few functions in `mopro` and we write `prove` in mopro with `ark-groth16`
    - What we use in mopro
       https://github.com/zkmopro/mopro/blob/e424efd0f6f9179aeda2fbf07885907b8c6d5dcb/mopro-ffi/src/circom/mod.rs#L16-L18
    - how we prove in circom
       https://github.com/zkmopro/mopro/blob/e424efd0f6f9179aeda2fbf07885907b8c6d5dcb/mopro-ffi/src/circom/mod.rs#L198-L206
    - We can maintain our own `circom-prover` to perform these functions.
 3. We have several adapters (now or in the future). We need to easily switch between adapters
    - We are doing research on GPU acceleration: https://github.com/zkmopro/gpu-acceleration
       If it improves the performance enough, we can switch from `ark-groth16` to e.g. `gpu-groth16`
    - We might have other witness generator in the future. e.g. [circom-witnesscalc](https://github.com/iden3/circom-witnesscalc) or [witnesscalc](https://github.com/0xPolygonID/witnesscalc)

We can maintain our own `circom-prover` for cross-platform solution, and we can publish our packages.
In the future, we can encourage ZK applications (e.g. Semaphore) to use the crate `circom-prover` instead of `circom-compat` for cross-platform proving.

## Details
Similar to `circom-compat`
It maintains the functions we need in mopro-ffi
https://github.com/zkmopro/mopro/blob/e424efd0f6f9179aeda2fbf07885907b8c6d5dcb/mopro-ffi/src/circom/mod.rs#L16-L18
https://github.com/zkmopro/mopro/blob/e424efd0f6f9179aeda2fbf07885907b8c6d5dcb/mopro-ffi/src/circom/serialization.rs#L3
It can also export functions called `prove`, `verify`
like it does in mopro-ffi
[...]

PR #303, opened 2025-01-20
Integrate c++ witness generator

alxkzmn:
The [wintesscalc fork](https://github.com/zkmopro/witnesscalc) is integrated via the [witnesscalc-adapter](https://github.com/zkmopro/witnesscalc_adapter.git). `witnesscalc-adapter` was inspired by `rust-witness` and has a similar API: [example 1](https://github.com/zkmopro/mopro/pull/303/files#diff-7eb9499f97f124e6cb1046405d397604e1d667ea965824971f532a4d9798fdadR14-R15), [example 2](https://github.com/zkmopro/mopro/pull/303/files#diff-3bf67bbb489f91e20c1656030cd85c82125905ff94a3d19bea97c271f1615011R7-R10).

## Learnings
- Linking C++ libraries has some peculiarities. For example, despite [giving the necessary directives](https://github.com/zkmopro/witnesscalc_adapter/blob/main/witnesscalc_adapter/src/lib.rs#L222) in the Rust code we still need to [explicitly state](https://github.com/zkmopro/mopro/pull/303/files#diff-36a86c5434d445643185208cb60101b6621a4d36c750c6477be3cd61c245535dR432) that we link the C++ standard library for iOS. However, having those directives in Rust is necessary to link the libraries for the tests running on the host. Overall, the [substantial changes](https://github.com/0xPolygonID/witnesscalc/compare/main...zkmopro:witnesscalc:main) to the `witnesscalc` fork are caused by the need to build binaries that are linkable with Rust instead of building the target platform artifacts like an xcode project.
## Changes
- the circuit zkey is expected to have the same filename as the witness genrator C++ source and the .dat file. This can be changed, but done like that for simplicity (see [here](https://github.com/zkmopro/mopro/pull/303/files#diff-fc55e54e611e20a9021cebf35830e8912bc0ced75f13d272e4ed9a5ebb7a9d69R137-R139)).
- NDK_PATH variable seemed to be unused (at least I don't have it set in my system but I was able to successfully compile for Android). However, witnesscalc needs to know the NDK location and expects `ANDROID_NDK` to be set. Therefore, [updated the docs](https://github.com/zkmopro/mopro/pull/303/files#diff-12e6b091146100567b6c5c796d0735f9ef7b5bc1d249084559f901493d5a6051R52). 
- Added [NASM installation](https://github.com/zkmopro/mopro/pull/303/files#diff-12e6b091146100567b6c5c796d0735f9ef7b5bc1d249084559f901493d5a6051R58-R74) as a requirement to the docs. NASM is necessary to compile the assembly sources of the finite field library that the witness generator is using.

## Remaining challenges
- `witnesscalc` does not support 32-bit architecture. Therefore, [I disabled them here](https://github.com/zkmopro/mopro/pull/303/files#diff-265feba39d20beb15fa4692ce2a52bad0e6bc3bad90482473e4058807e9beddcR48-R54). There are some possible ways of dealing with it, e.g., using rust-witness if 32-bit is a must, or potentially adding the 32-bit implementation of the finite field arithmetic. [Here's the WIP branch](https://github.com/zkmopro/witnesscalc/compare/main...zkmopro:witnesscalc:i686) for 32-bit support, it won't compile because the "generic" implementation of Fr [here](https://github.com/zkmopro/witnesscalc/blob/main/build/fr_generic.cpp) and [here](https://github.com/zkmopro/witnesscalc/blob/main/build/fr_raw_generic.cpp) is explicitly using some 64-bit types.
- `witnesscalc` does not support the `bls12381` curve because the finite field implementation has [hardcoded constants for bn128](https://github.com/zkmopro/witnesscalc/blob/main/build/fr_generic.cpp#L5). This is potentially solvable as even the Circom itself generates the `fr.cpp` as part of the C++ output with the constants corresponding to the selected field. The challenge is in making the Fr C++ implementation "dynamic" according to the field (should we template it out in `witnesscalc`? or have two hardocded options for bn and bls? is there a way to provide the whole `fr.cpp` output by Circom as an input to mopro build pipeline?). In relation to this, an [extra test was added](https://github.com/zkmopro/mopro/pull/303/files#diff-fc55e54e611e20a9021cebf35830e8912bc0ced75f13d272e4ed9a5ebb7a9d69R512-R550) to test the `hashbench` circuit with witnesscalc but for `bn128` instead of `bls`.


chancehudson:
I think we should stop pursuing this approach of binding c++ into rust into mobile. We're trying to wrap a complex build system in a complex build system and then give that to the user. I think instead we should maintain a rapidsnark fork that is intended to be integrated into mobile projects. (sorta like what alex said in discord a while back, c++ only no rust)

I think even if we slog through these errors we're going to introduce binary bloat and possibly runtime overhead by wrapping the system as we're doing here. I also think it will not be compatible with other people's build machines - there are too many variables: os, architecture, rust version, c++ stdlib, compiler, ios version, android version, etc.

Mopro is glue code for rust projects. I thought it was _maybe_ possible to wrap c++ stuff as well, but now i think it's not really. Maybe it would be fine if it was just c++, but it's c++ linking to large libraries and assembly implementations.

The frustrating part is, linking rapidsnark directly into an iOS project isn't that complicated. If you take out the intermediate rust build step it's not hard to build the rapidsnark static libraries and put those in an iOS project. I really question the value of projects like mopro that burn hundreds/thousands of developer hours to prevent other developers from learning the tools they're using, without providing other value/features.

vivianjeng:
I think we are almost there
to wrap up witnesscalc in Rust
and we can benefit from the performance in Rust
```sh
➜  tests git:(ci) ✗ cargo run --package tests --bin tests --release
   Compiling tests v0.1.0 (/Users/zhengyawen/Documents/GitHub/witnesscalc_adapter/tests)
    Finished `release` profile [optimized] target(s) in 0.41s
     Running `/Users/zhengyawen/Documents/GitHub/witnesscalc_adapter/target/release/tests`
Generating witness for circuit keccak256_256_test
Time taken: 71.483584ms
```
You can check the benchmark here
https://zkmopro.org/docs/performance#macos

I think it still worth exploring
in the future, we might have applications like Semaphore
and we don't want to separate the functions between Rust and the mobile native code.
e.g. 
I create a `semaphore-rs` but it uses `rust-witness`+`ark-works` to generate proofs (or maybe not generate proofs, but circuit inputs)
but in mobile native, it uses different functions `witnesscalc`+`rapidsnark`
I think it will be less convenient for developers

but I am thinking we move the `witnesscalc` to one of the features in `circom-prover`
see: #299 and https://github.com/zkmopro/circom-prover
by default it can just use `rust-witness`
and users can also use 
```toml
circom-prover = {git = "https://github.com/zkmopro/circom-prover", features = ["witnesscalc"]}
```
to enable `witnesscalc` feature
(the same as for `circom-witnesscalc`)

chancehudson:
I think we should continue to integrate into rust, but not try to integrate through to mobile. Instead the recommended path to get rapidsnark on mobile could be linking to the c++ library itself.

@alxkzmn what do you think? Those ci errors look nasty and i just learned today my static library linking approach seems to not be working across different c++ standard library implementations and the symbols seemed to be getting garbled by the rust intermediate linking step 😔

I can try reworking it though to compile from source i guess which should be easier given your work on witnesscalc, if you think it's feasible.

alxkzmn:
@chancehudson I think it may be worth trying to compile from sources similarly to witnesscalc, yes. It seems like the CI errors are a minor issue now with known solutions. Simply speaking, we're too close to make it work to scrap it. Especially if the introduced overhead turns out to be negligible. 

I agree that wrapping C++ into Rust introduces a lot of complexity, but I also agree with @vivianjeng regarding potential integrations with Rust codebases like Semaphore. Also, linking native mobile code to precompiled binaries definitely has the least performance overhead but I'd argue that it introduces the most DevEx overhead. The advantage of mopro is that in the end it's a single library, not a separate witgen+prover that require separate build and integration into mobile. It would be nice to base the decision on the feedback from the real mopro users though. Are they willing to build witgen+prover and learn about FFI on mobile or they prefer simplicity that mopro would provide?

chancehudson:
alright i'll stop my complaining and make it work

vivianjeng:
@chancehudson you can check how Alex integrates with `witnesscalc`
https://github.com/zkmopro/witnesscalc_adapter
I think it would be similar to `rapidsnark` (not sure)

vivianjeng:
this will be integrated in circom-prover #299

PR #304, opened 2025-01-22
Feat/circom prover

### LINKED ISSUE #299: `circom-prover` spec

KimiWu123:
close #299


circom-prover/src/witness.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+pub enum WitnessLib {
+    WitnessCalc,
+    RustWitness,
+}
+
+/// To create witness functions corresponding to different witness generation libs.
+#[macro_export]
+macro_rules! create_witness_fn {
```
chancehudson, line 21: I think this macro is a bad abstraction. Why do we want to hide which adapter is being used to create the witness function? e.g. why can't we use `rust_witness` and `witnesscalc_adapter` directly? Adding this adds a layer of complexity, what is the justification?

circom-prover/src/witness.rs
```diff
@@ -0,0 +1,63 @@
...
+        compile_error!(concat!("Unknown macro type: ", stringify!($other)));
+    };
+}
+
+pub fn generate_witness(
+    witness_fn: WitnessFn,
+    inputs: HashMap<String, Vec<String>>,
+    dat_path: String,
+) -> Vec<BigUint> {
+    std::thread::spawn(move || {
```
chancehudson, line 38: This thread is a noop. This `generate_witness` implementation creates a background thread, then blocks the main thread waiting for the background thread (e.g. immediately joins the thread below).

chancehudson (commented):
Could i get a concrete rationale for creating a new package in the monorepo instead of breaking the ark logic into a separate file instead? This adds a layer of complexity and as such i'd expect concrete upside. As it stands i don't see the upside and am against adding a new package.

chancehudson:
ah i see, the intention is to publish the package on crates.io?

KimiWu123:
> Could i get a concrete rationale for creating a new package in the monorepo instead of breaking the ark logic into a separate file instead? This adds a layer of complexity and as such i'd expect concrete upside. As it stands i don't see the upside and am against adding a new package.

I think Vivian's intention is 
> We can maintain our own circom-prover for cross-platform solution, and we can publish our packages.
In the future, we can encourage ZK applications (e.g. Semaphore) to use the crate circom-prover instead of circom-compat for cross-platform proving.

, want to address 
> We have several adapters (now or in the future). We need to easily switch between adapters

and also to simply the dependencies of `mopro`, [here](https://github.com/zkmopro/mopro/blob/main/mopro-ffi/Cargo.toml#L20C1-L34C2).

At the beginning, Vivian proposed to have a separated [repo](https://github.com/zkmopro/circom-prover) to address above issues. However, I suggested to put the logic inside `mopro` repo and have another `crate`. The reason why I suggest that is because it'll be getting harder and harder to maintain if we separate our code in so many different repos. And it's also hard for newcomers to get whole pictures of `mopro`.  

I would love to hear your feedback on how to fix #299 and make the codebase clean! 😄

vivianjeng:
@chancehudson Yes
to be clarified
It will be `circom-prover` (https://github.com/zkmopro/mopro/issues/299) in the future
(will be here https://github.com/zkmopro/circom-prover)
But for current stage, it would be convenient to put in monorepo like this

and I think the `rapidsnark` can also be in part of `circom-prover`
like we can set different features to enable different adapter
e.g. `features = ["ark-works"]` or `features = ["rapidsnark"]`
how do you think? related to this PR https://github.com/zkmopro/mopro/pull/305

and I think the `rapidsnark-adpater` can also be a crate like [`witnesscalc-adpater`](https://github.com/zkmopro/witnesscalc_adapter)
I think they are quite independent to mopro or `circom-prover`

circom-prover/src/witness.rs
KimiWu123: Nice catch! I forgot to handle this. fixed in [b6d39a6](https://github.com/zkmopro/mopro/pull/304/commits/b6d39a6ba3acaddc4a0e7186b695e64ca208708a)

circom-prover/src/prover.rs
```diff
@@ -0,0 +1,39 @@
...
+pub fn prove(
+    lib: ProofLib,
+    zkey_path: String,
+    witnesses: JoinHandle<Vec<BigUint>>,
+) -> Result<CircomProof> {
+    match lib {
+        ProofLib::Arkworks => arkworks::generate_circom_proof(zkey_path, witnesses),
+        ProofLib::RapidSnark => panic!("Not supported yet."),
+    }
+}
```
vivianjeng, line 27: I think it can be controlled by feature flags
the dependencies will be built when users enable one feature/or two features

vivianjeng: the same as the witness calculator

circom-prover/src/witness.rs
KimiWu123: removed in [b22be2c](https://github.com/zkmopro/mopro/pull/304/commits/b22be2cff51e3db57501cb480fbdd60bba92e727)

KimiWu123: removed in [d47fd38](https://github.com/zkmopro/mopro/pull/304/commits/d47fd38ce6edda89ffbc03479b6abb2b505330d7)

circom-prover/src/prover.rs
KimiWu123: fixed in [265abb6](https://github.com/zkmopro/mopro/pull/304/commits/265abb65315544aa49003ca2e878f62fa9037f2c)

vivianjeng (commented):
I think the more ideal implementation is to have like
```
witness/
   witnesscalc.rs
   rust_witness.rs
   mod.rs
```
and we enable each by
```rs
#[cfg(feature = "witnesscalc")]
mod witnesscalc;
#[cfg(feature = "rust-witness")]
mod rust_witness;
```
so we don't have to set many feature flags
and what if people set two different witness generators at the same time?
e.g. `rustwitness` and `witnesscalc`

it looks like it will execute the same operation twice?
```rs
            #[cfg(feature = "witnesscalc")]
            WitnessFn::WitnessCalc(wit_fn) => wit_fn(bigint_inputs, _dat_path.as_str()),
            #[cfg(feature = "rustwitness")]
            WitnessFn::RustWitness(wit_fn) => wit_fn(bigint_inputs),
```

I actually prefer the way if we can enable both features at the same time
(though it might not make sense)
but we need to design the functions
like `generate_witness_rust_witness`, `generate_witness_witnesscalc`
and init the prover like
`let circom_prover = CircomProver(wtns_fn, proof_fn);`

or we need to check if one feature is enabled, and another should not be executed

KimiWu123:
> I think the more ideal implementation is to have like
> 
> ```
> witness/
>    witnesscalc.rs
>    rust_witness.rs
>    mod.rs
> ```
> 
> and we enable each by
> 
> ```rust
> #[cfg(feature = "witnesscalc")]
> mod witnesscalc;
> #[cfg(feature = "rust-witness")]
> mod rust_witness;
> ```

It's a good idea, but I realized there is no much in witnesscalc.rs or rust_witness.rs. It seems a little over design for me and that's why I kept two different witness generation functions in the same file.

> so we don't have to set many feature flags and what if people set two different witness generators at the same time? e.g. `rustwitness` and `witnesscalc`
> 
> it looks like it will execute the same operation twice?
> 
> ```rust
>             #[cfg(feature = "witnesscalc")]
>             WitnessFn::WitnessCalc(wit_fn) => wit_fn(bigint_inputs, _dat_path.as_str()),
>             #[cfg(feature = "rustwitness")]
>             WitnessFn::RustWitness(wit_fn) => wit_fn(bigint_inputs),
> ```

No, it won't execute the same operation twice. `wit_fn` is passed from outside, it's the first argument of `generate_witness`. Callers can call which witness generation function as they want. For example, if both features were set, the code would be 
```rust 
             WitnessFn::WitnessCalc(wit_fn) => wit_fn(bigint_inputs, _dat_path.as_str()),
             WitnessFn::RustWitness(wit_fn) => wit_fn(bigint_inputs),
```
Just like a normal `match` case.

KimiWu123:
> let circom_prover = CircomProver(wtns_fn, proof_fn);

I was considering this design as well, but it seems not really necessary to do in this way. I think current design is more lightweight, people can call any functions as they want, not have to init a `CircomProver` first. It better to have multiple choices for witness and proving in this kind of desgin, for example,
```rust
let a = CircomProver(rust_witness, arkworks);
let b = CircomProver(witness_calc, arkworks);
```
which means it better not to use `cfg(feature)` to control the output of `CircomProver`, supports all features at a time. Otherwise, it feel strange that I can use `rust_witness` in this release/package but can't in another release. Not sure if it's clear enough.

The only benefit I can image is that we don't need to pass `witness`/`witness_thread` around, just a single `prove` can handle everything.

test-e2e/src/lib.rs
```diff
@@ -7,12 +7,12 @@ mopro_ffi::app!();
 rust_witness::witness!(multiplier2);
 rust_witness::witness!(multiplier2bls);
 rust_witness::witness!(keccak256256test);
-rust_witness::witness!(hashbenchbls);
+rust_witness::witness!(hashbench_bls);
```
vivianjeng, line 10: I remember it doesn't accept `_` and `-`
though there is no test on `hashbenchbls`
maybe you can just delete this line
```suggestion
```

circom-prover/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+pub mod prover;
+pub mod witness;
+
+// pub use rust_witness::transpile::transpile_wasm;
```
vivianjeng, line 4: this can be deleted as well?
```suggestion
```

vivianjeng (approved):
LGTM
Just some lines can be removed

PR #305, opened 2025-01-22
Rapidsnark support

chancehudson:
This PR adds rapidsnark proving support.

The rapidsnark library and dependencies are prebuilt for each architecture to avoid developers having to fully compile rapidsnark, which requires system dependencies. 

This PR currently disables 32 bit android support. 32 bit support was removed in android 14, which was released in 2023. We should discuss this more, but I believe we should disable 32 bit support.

Rapidsnark is compiled from the fork [here](https://github.com/chancehudson/rapidsnark), which adds externally exported functions for building and verifying proofs.

TODO:
- remove fat libraries
- document exporting the libraries
- possibly add a script to auto build all rapidsnark static libraries
- double check that rapidsnark libraries are built in release mode
- compile rapidsnark for `armv7` (potentially fall back to ark instead?)


vivianjeng:
this will be integrated in circom-prover #299

PR #308, opened 2025-01-25
Refactor/circom prover

KimiWu123:
Tried to address https://github.com/zkmopro/mopro/pull/304#pullrequestreview-2571868681.

---

Hi @vivianjeng , 

>and init the prover like
`let circom_prover = CircomProver(wtns_fn, proof_fn);`

I tried implement above idea, but it's hard to pass the `circom_prover` object from [generate_circom_proof_wtns](https://github.com/zkmopro/mopro/blob/main/mopro-ffi/src/circom/mod.rs#L26) to [verify_circom_proof](https://github.com/zkmopro/mopro/blob/main/mopro-ffi/src/circom/mod.rs#L41). So, it can't achieve what you expected and that's the reason why I only implemeted static functions in this PR. Does this refactor make sense to you?


circom-prover/src/witness.rs
```diff
@@ -18,7 +19,7 @@ pub enum WitnessFn {
 pub fn generate_witness(
     witness_fn: WitnessFn,
     inputs: HashMap<String, Vec<String>>,
-    _dat_path: String,
+    _zkey_path: String,
```
vivianjeng, line 22: I think we should keep this `_dat_path`.
It is quite common that the zkey name is different from dat name

I am also thinking to make an improvement in `witnesscalc-adapter` to let it find .dat automatically
so it doesn't have to take `.dat` as a function input

vivianjeng (commented):
overall it looks good to me
when the `dat_path` is fixed, we can merge this PR

circom-prover/src/witness.rs
vivianjeng: I created a PR here
https://github.com/zkmopro/witnesscalc_adapter/pull/5
but it might break the current mopro
I think we can merge this after this PR is merged

KimiWu123: fixed in [183fa83](https://github.com/zkmopro/mopro/pull/308/commits/183fa83bd620e770a6a5dbc477f62cb017ee92ed)

Issue #306, opened 2025-01-22
Enhance Mopro for Generic Workflows


ElusAegis:
# Problem
Mopro's current design assumes packages are primarily used for Zero-Knowledge Proofs (ZKP), enforcing default `prove` and `verify` functions. While useful for ZKP-specific tasks, this limits its utility for projects requiring generic or fully custom workflows.

## Details
To address these issues, the following solutions are proposed:

1. **Support for Custom Workflows:**
   - Make `prove` and `verify` functions optional.
   - Enable developers to define custom Rust functions with a single procedural macro, such as `#[uniffi::export]` or a custom equivalent.

2. **Move Beyond UDL Files:**
   - Transition from UDL files to procedural macros to simplify integrations and offer greater flexibility.

## Acceptance Criteria
- `prove` and `verify` functions are optional, and developers can define fully custom workflows.
- Procedural macros replace UDL files, simplifying integration and improving flexibility.

vivianjeng:
Thank you for the suggestion 🙏🏻
I will consider this more and discuss with our team members

Simply summarize this issue's solution and the current implementation:
**Current implementation:**
- Define functions like
   ```rs
   pub fn generate_circom_proof(...) {
      #[cfg(feature = "circom")]
      // implement prover
      #[cfg(not(feature = "circom"))]
      panic!("Circom is not enabled in this build. Please pass `circom` feature to `mopro-ffi` to enable Circom.")
   }
   ```
- Generate UDL file like
   ```rs
   [Throws=MoproError]
   GenerateProofResult generate_circom_proof(string zkey_path, record<string, sequence<string>> circuit_inputs);
   ```
- `build.rs` file
   ```rs
   uniffi::generate_scaffolding(udl_path.to_str().unwrap()).unwrap();
   ```

**Procedural macros**
- Define functions like
   ```rs
   // Option 1: make `generate_circom_proof` optional (suggested in the issue)
   #[cfg(feature = "circom")]
   #[uniffi::export]
   pub fn generate_circom_proof(...) {
      // implement prover
   }

   // Option 2: keep the `generate_circom_proof` interface
   #[uniffi::export]
   pub fn generate_circom_proof(...) {
      #[cfg(feature = "circom")]
      // implement prover
      #[cfg(not(feature = "circom"))]
      panic!("Circom is not enabled in this build. Please pass `circom` feature to `mopro-ffi` to enable Circom.")
   }
   ```
- `lib.rs` file
   ```rs
[...]

PR #312, opened 2025-02-03
feat/circom prover removes ark-circom dep.

KimiWu123:
This PR is the second part of #299 , removing `ark-circom` dependency. 

#### Details:
Copied functions we need from https://github.com/zkmopro/circom-compat/tree/wasm-delete to our repo.


circom-prover/src/prover/serialization.rs
```diff
@@ -39,40 +36,13 @@ pub fn deserialize_inputs<T: Pairing>(data: Vec<u8>) -> SerializableInputs<T> {
     SerializableInputs::deserialize_uncompressed(&mut &data[..]).expect("Deserialization failed")
 }
 
-// Convert proof to U256-tuples as expected by the Solidity Groth16 Verifier
```
KimiWu123, line 42: These two functions were moved to `mopro-ffi/circom`

KimiWu123:
@vivianjeng , this PR is ready for review! 

I decided to keep serialization functions in `circom-prover` because I think it's easier to pass native types (`Vec<u8>`) cross modules, don't need to handle generic types (`Proof<T: Pairing>`).

mopro-ffi/src/lib.rs
```diff
@@ -129,9 +129,8 @@ macro_rules! app {
         use circom_prover::{
             prover::{CircomProof, ProofLib},
             witness::WitnessFn,
-            ProofCalldata, G1, G2,
         };
```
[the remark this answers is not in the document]
KimiWu123: Nice catch, but we still need to keep `WitnessFn`. It is used in `test-e2e/src/lib.rs`, like the following,
```
    ("multiplier2_final.zkey", WitnessFn::RustWitness(multiplier2_witness))
}
```

fixed in [7d42ac1](https://github.com/zkmopro/mopro/pull/312/commits/7d42ac161b0d3ce9f90314e39edab6cba0003dd9)

mopro-ffi/src/lib.rs
```diff
@@ -126,12 +126,8 @@ pub struct GenerateProofResult {
 macro_rules! app {
     () => {
         // These are mandatory imports for the uniffi to pick them up and match with UDL
-        use circom_prover::{
-            prover::{CircomProof, ProofLib},
-            witness::WitnessFn,
-            ProofCalldata, G1, G2,
-        };
-        use mopro_ffi::{GenerateProofResult, MoproError};
+        use circom_prover::witness::WitnessFn;
```
vivianjeng, line 129: I am still concern about the import
but we can fix this by this issue 
if the circom feature is disabled
https://github.com/zkmopro/mopro/blob/7d42ac161b0d3ce9f90314e39edab6cba0003dd9/mopro-ffi/Cargo.toml#L19-L20
Apps will not find the `circom_prover` dependency

but it can be fixed by this issue
https://github.com/zkmopro/mopro/issues/306

vivianjeng (approved):
LGTM

mopro-ffi/src/lib.rs
KimiWu123: Good point, I didn't think about that. I submitted [d3234fb](https://github.com/zkmopro/mopro/pull/312/commits/d3234fb605ab3b1ed3c57e124d811ac2206c4096) to fix this. The downside of this solution is caller (`test-e2e` needs to define `circom` feature in toml file). The other solution is to implement an empty struct like `circom_app!` and `halo2_app!`.

KimiWu123: @vivianjeng , I just realized, our current implementation also has this issue. For example, if we disable `circom` feature, it wll cause an error that `set_circom_circuits` can't be found. So, I think it's a good idea to address this issue in #306 as you mentioned.

PR #321, opened 2025-02-12
Applied uniffi macro instead of using udl file.

sifnoc:
Todo
- [ ] Fix CI error

### DIFF
--- a/mopro-ffi/src/app_config/android.rs
+++ b/mopro-ffi/src/app_config/android.rs
@@ -2,9 +2,6 @@ use std::fs;
 use std::path::Path;
 use std::process::Command;
 
-use uniffi::generate_bindings;
-use uniffi::KotlinBindingGenerator;
-
 use super::cleanup_tmp_local;
 use super::constants::{
     AndroidArch, Mode, ARCH_ARM_64_V8, ARCH_ARM_V7_ABI, ARCH_I686, ARCH_X86_64, ENV_ANDROID_ARCHS,
@@ -49,16 +46,17 @@ pub fn build() {
         build_for_arch(arch, &build_dir, &bindings_out, mode);
     }
 
-    generate_bindings(
-        (manifest_dir + "/src/mopro.udl").as_str().into(),
-        None,
-        KotlinBindingGenerator,
-        Some(bindings_out.to_str().unwrap().into()),
-        None,
-        None,
-        false,
-    )
-    .expect("Failed to generate bindings");
+    //TODO: remove this
+    // generate_bindings(
+    //     (manifest_dir + "/src/mopro.udl").as_str().into(),
+    //     None,
+    //     KotlinBindingGenerator,
+    //     Some(bindings_out.to_str().unwrap().into()),
+    //     None,
+    //     None,
+    //     false,
+    // )
+    // .expect("Failed to generate bindings");
 
     move_bindings(&bindings_out, &bindings_dest);
     cleanup_tmp_local(&build_dir);
--- a/mopro-ffi/src/app_config/android.rs
+++ b/mopro-ffi/src/app_config/android.rs
@@ -46,17 +46,31 @@ pub fn build() {
         build_for_arch(arch, &build_dir, &bindings_out, mode);
     }
 
-    //TODO: remove this
-    // generate_bindings(
-    //     (manifest_dir + "/src/mopro.udl").as_str().into(),
-    //     None,
-    //     KotlinBindingGenerator,
-    //     Some(bindings_out.to_str().unwrap().into()),
-    //     None,
-    //     None,
-    //     false,
-    // )
-    // .expect("Failed to generate bindings");
+    // Uniffi proc-macro require compiled library file
+    Command::new("cargo")
+        .args([
+            "run",
+            "--bin",
+            "uniffi-bindgen",
+            "generate",
+            "--library",
+            // Compiled lib out dir
+            build_dir
+                .join(if mode == Mode::Release {
+                    "release"
+                } else {
+                    "debug"
+                })
+                .join("deps/libmopro_ffi.so")
+                .to_str()
+                .expect("Invalid static library path"),
+            "--language",
+            "kotlin",
+            "--out-dir",
+            bindings_out.to_str().expect("Invalid output directory"),
+        ])
+        .status()
+        .expect("Failed to execute uniffi-bindgen command");
 
     move_bindings(&bindings_out, &bindings_dest);
     cleanup_tmp_local(&build_dir);
--- a/mopro-ffi/src/app_config/ios.rs
+++ b/mopro-ffi/src/app_config/ios.rs
@@ -94,29 +94,31 @@ pub fn build() {
         .map(|v| build_combined_archs(v))
         .collect();
 
-    // TODO: Should i generate all
-    let generate_binding_cmds = Command::new("cargo")
+    // Uniffi proc-macro require compiled library file
+    Command::new("cargo")
         .args([
             "run",
             "--bin",
             "uniffi-bindgen",
             "generate",
-            "--lib-file",
-            out_lib_paths[0]
+            "--library",
+            // Compiled lib out dir
+            build_dir_path
+                .join(if mode == Mode::Release {
+                    "release"
+                } else {
+                    "debug"
+                })
+                .join("deps/libmopro_ffi.so")
                 .to_str()
                 .expect("Invalid static library path"),
             "--language",
             "swift",
             "--out-dir",
-            swift_bindings_dir
-                .to_str()
-                .expect("Invalid output directory"),
+            bindings_out.to_str().expect("Invalid output directory"),
         ])
         .status()
         .expect("Failed to execute uniffi-bindgen command");
-
-    // TODO: remove cli
-    println!("generate_binding_cmds: {:?}", generate_binding_cmds);
 
     fs::rename(
         swift_bindings_dir.join("mopro.swift"),
--- a/mopro-ffi/src/app_config/ios.rs
+++ b/mopro-ffi/src/app_config/ios.rs
@@ -2,9 +2,6 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-use uniffi::generate_bindings;
-use uniffi::SwiftBindingGenerator;
-
 use super::cleanup_tmp_local;
 use super::constants::{IosArch, Mode, ARCH_ARM_64, ARCH_X86_64, ENV_CONFIG, ENV_IOS_ARCHS};
 use super::install_arch;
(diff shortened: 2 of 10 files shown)


sifnoc:
Notes for recent changes:
- Use the re-export methods introduced with feature flags in UniFFI v0.29 instead of adding uniffi-* dependencies
- Mopro macros requires return types-`GenerateProofResult`, `MoproError` due to Uniffi limitations
- The CLI has been updated to work without UDL
- Removed the UDL introduction from the docs(Including a blog post)
- Commented out the failing test `test-e2e` on CI – this will be handled in the next PR (@vivi)

mopro-ffi/src/lib.rs
```diff
@@ -145,13 +142,83 @@ pub struct ProofCalldata {
...
+                match err {
+                    mopro_ffi::MoproError::CircomError(e) => Self::CircomError(e),
+                    mopro_ffi::MoproError::Halo2Error(e) => Self::Halo2Error(e),
+                    _ => panic!("Unhandled error type: {}", err),
+                }
+            }
+        }
+
+        #[derive(Debug, Clone, uniffi::Record)]
+        pub struct GenerateProofResult {
```
KimiWu123, line 168: I still don't understand why we need to declare `GenerateProofResult` and `ProofCalldata` here. 

Do you mean if we declare in this way, 
```rust 
macro_rules! circom_app {
    () => {
        use mopro_ffi::witness::WitnessFn;
        fn generate_circom_proof(
            zkey_path: String,
            inputs: std::collections::HashMap<String, Vec<String>>,
        ) -> Result<mopro_ffi::GenerateProofResult, mopro_ffi::MoproError> {
        ....
        }
}
```
it doesn't work because of what you mentioned in the description: 
> However, uniffi does not work correctly when calling internal functions that are not processed by the uniffi proc-macro. In other words, even internal functions require their return types to have uniffi declarations.

and the reason behind is the uniffi limitation you commented above, https://github.com/mozilla/uniffi-rs/issues/2257? Do I understand it correctly? 

One more question, if I want to reproduce this error, which test cases I should run? Or how should I reproduce it?

KimiWu123 (approved):
Looks awesome! Left a comment, want to know more about the Uniffi limitation you mentioned.
The code looks nice and clean now. However, it seems need more knowledge to write and export functions from Rust code.

cli/src/template/init/src/lib.rs
```diff
@@ -1,5 +1,7 @@
-// Here we're calling a macro exported by uniffi. This macro will
-// write some functions and bind them to the uniffi UDL file. These
+use circom_prover::witness::WitnessFn;
```
vivianjeng, line 1: maybe we should put this in `// CIRCOM_TEMPLATE`
so if `circom` feature is not activated, it won't be an error

docs/blog/2025-01-06-2024-retro.md
```diff
@@ -44,26 +44,22 @@ Please refer to the following resources to learn how to use Mopro for building W
...
 
 ```rust
+#[uniffi::export]
 pub fn poseidon(input: Vec<u8>) -> Vec<u8>{
     // Poseidon hash implementation
 }
 ```
 
-Next, they define the corresponding API in the UDL file. Please refer to [uniffi-rs](https://mozilla.github.io/uniffi-rs/latest/udl_file_spec.html) to learn more about the UDL file format.
+By annotating your function with #[uniffi::export], UniFFI automatically declares the appropriate FFI type and generates the necessary bindings for Swift, Kotlin.
```
vivianjeng, line 58:
```suggestion
By annotating your function with `#[uniffi::export]`, UniFFI automatically declares the appropriate FFI type and generates the necessary bindings for Swift, Kotlin. 
```

docs/docs/setup/rust-setup.md
```diff
@@ -220,30 +216,26 @@ Download example SRS and key files :
...
-    // Feel free to modify it to suit your needs.
-    let udl_path = Path::new("src/mopro.udl");
-    if !udl_path.exists() {
-        std::fs::write(udl_path, mopro_ffi::app_config::UDL).expect("Failed to write UDL");
-    }
-    // Finally initialize uniffi and build the scaffolding into the
-    // rust binary
-    uniffi::generate_scaffolding(udl_path.to_str().unwrap()).unwrap();
+    // For removing warnings
+    println!("cargo::rustc-check-cfg=cfg(disable_uniffi_export)");
```
vivianjeng, line 226: can we just put it in `mopro_ffi` `build.rs`?
and I think if we will use `rust_witness::witness!(multiplier2);`
we should add
```rs
rust_witness::transpile::transpile_wasm("../test-vectors/circom".to_string());
```

mopro-ffi/src/app_config/android.rs
```diff
@@ -116,3 +114,25 @@ fn move_bindings(bindings_out: &Path, bindings_dest: &Path) {
     }
     fs::rename(bindings_out, bindings_dest).expect("Failed to move bindings into place");
 }
+
+fn generate_android_bindings(dylib_path: &Path, binding_dir: &Path) -> Result<(), Error> {
```
vivianjeng, line 118: I think with `use anyhow::Result;`
we don't have to define `Error`
just use `Result<()>`?

mopro-ffi/src/app_config/ios.rs
```diff
@@ -201,3 +202,25 @@ fn rename_module_maps_recursively(bindings_out: &PathBuf) {
         }
     }
 }
+
+fn generate_ios_bindings(dylib_path: &Path, binding_dir: &Path) -> Result<(), Error> {
```
vivianjeng, line 206: same here

mopro-ffi/src/lib.rs
sifnoc: Technically, I encountered two issues.

The first issue was that the return and error types—specifically `GenerateProofResult` and `MoproError`—in the `generate_circom_proof` method exposed via UniFFI need to be processed by the UniFFI proc-macro. However, `mopro_ffi::MoproError` was not annotated with the UniFFI proc-macro. To work around this, I added a type conversion function, for example:

```rust
impl From<mopro_ffi::GenerateProofResult> for GenerateProofResult {
    fn from(result: mopro_ffi::GenerateProofResult) -> Self {
        Self {
            proof: result.proof,
            inputs: result.inputs,
        }
    }
}
```
This relates to what I mentioned in my initial comment.

The second issue was that the UniFFI proc-macro could not locate the FFI-exposed type when it was declared in an external crate, which is tied to the issue discussed in #2257. To resolve this, I declared the types within the macro and used them in the same crate for test-e2e, rather than in "mopro-ffi"

I hope this explanation helps clarify the situation.

> One more question, if I want to reproduce this error, which test cases I should run? Or how should I reproduce it?

I lost commit has the error `unknown throw type: Some(External`. but It happened in the `test-e2e` while executing `cargo run --bin ios` or `cargo run --bin android`.

docs/docs/setup/rust-setup.md
sifnoc: It's only for turning off warning like:
```
warning: unexpected `cfg` condition name: `disable_uniffi_export`
 --> test-e2e/src/lib.rs:5:1
  |
5 | mopro_ffi::app!();
  | ^^^^^^^^^^^^^^^^^
  |
  = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
  = help: try referring to `mopro_ffi::halo2_app` crate for guidance on how handle this unexpected cfg
  = help: the macro `mopro_ffi::halo2_app` may come from an old version of the `mopro_ffi` crate, try updating your dependency with `cargo update -p mopro_ffi`
  = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
  = note: this warning originates in the macro `mopro_ffi::halo2_app` which comes from the expansion of the macro `mopro_ffi::app` (in Nightly builds, run with -Z macro-backtrace for more info)
  ```
  And also It's already declared in the "mopro-ffi/build.rs"

Comment (vivianjeng) [reply docs/docs/setup/rust-setup.md:226]: it should be inherited from `mopro_ffi`?
if adding `mopro-ffi` in `build-dependencie`

Comment (vivianjeng) [reply docs/docs/setup/rust-setup.md:226]: it works for me if I removed `println!("cargo::rustc-check-cfg=cfg(disable_uniffi_export)");` in `test-e2e`

Comment (sifnoc) [reply docs/docs/setup/rust-setup.md:226]: > it works for me if I removed `println!("cargo::rustc-check-cfg=cfg(disable_uniffi_export)");` in `test-e2e`

It looks you couldn't see any warning like one above. I commit without the cfg statement in the `build.rs` in test-e2e.
Please check this errors on `clippy` test on CI: https://github.com/sifnoc/mopro/actions/runs/13627598826/job/38088087293#step:3:786

**Comment by vivianjeng** [comment]
I tried to disabled `halo2` or `circom` feature in e2e
but it throws the error
```sh
error: no rules expected `GenerateProofResult`
 --> test-e2e/src/lib.rs:5:1
  |
5 | mopro_ffi::app!();
  | ^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
  |
  = note: while trying to match end of macro
  = note: this error originates in the macro `mopro_ffi::app` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `test-e2e` (lib) due to 1 previous error
```

and I think we should handle if either feature is not activated in `mopro-example-app`
and since we keep the function if the feature is not activated
https://github.com/zkmopro/mopro/blob/41a083106fc43cda3d8583eaa99cdeca9fae7270/mopro-ffi/src/lib.rs#L20-L29
we should make it also export the function too?

sifnoc:
> I tried to disabled halo2 or circom feature in e2e

Would you give me more context for doing this?

> https://github.com/zkmopro/mopro/blob/41a083106fc43cda3d8583eaa99cdeca9fae7270/mopro-ffi/src/lib.rs#L20-L29
> 
> 
> we should make it also export the function too?

It would not happen in `mopro-example-app` that created by mopro cli. It because the `set_halo2_circuits!` or `set_circom_circuits!`statements would be included with the feature flag. For example, If I only select `circom` while initializing mopro example with the cli. the `src/lib.rs` would be like one below:

```rust
// Mopro-example-app/src/lib.rs
// Here we're calling a macro exported with Uniffi. This macro will
// write some functions and bind them to FFI type. These
// functions will invoke the `get_circom_wtns_fn` generated below.
mopro_ffi::app!();

// --- Circom Example of setting up multiplier2 circuit ---
use circom_prover::witness::WitnessFn;

rust_witness::witness!(multiplier2);

    ("multiplier2_final.zkey", WitnessFn::RustWitness(multiplier2_witness))
}
```
In `halo2` flag on case, also same.

sifnoc:
> we should make it also export the function too?

Exposed the stub macros - `halo2_app` and `circom_app` with macro variables. It worked well with one single feature flag(`circom` or `halo2`) in the cli.

.github/workflows/build-and-test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Run Clippy
-        run: cargo clippy --all-targets --all-features
+        run: DISABLE_UNIFFI_EXPORT=1 cargo clippy --all-targets --all-features
```
vivianjeng, line 24: I think you can put it in `.cargo/config.toml`
so that other users won't need to set this env

cli/src/template/init/build.rs
```diff
@@ -1,15 +1,6 @@
-use std::path::Path;
-
 fn main() {
+    // For removing warnings
+    println!("cargo::rustc-check-cfg=cfg(disable_uniffi_export)");
```
vivianjeng, line 3: It should be removed like the `build.rs` in `test-e2e`?

docs/docs/setup/rust-setup.md
```diff
@@ -69,27 +71,26 @@ Download example multiplier2 wasm and zkey here:
 Now we need to add four rust files. First we'll add a `build.rs` in the main project folder. This file should contain the following:
 
 ```rust
-use std::path::Path;
 fn main() {
-    // We're going to transpile the wasm witness generators to C
-    // Change this to where you put your zkeys and wasm files
+    fn main() {
+    // For removing warnings
+    println!("cargo::rustc-check-cfg=cfg(disable_uniffi_export)");
```
vivianjeng, line 77: same here

docs/docs/setup/rust-setup.md
vivianjeng: is it fixed by the flag `#![allow(unexpected_cfgs)]`?

vivianjeng (approved):
LGTM
Thank you!

.github/workflows/build-and-test.yml
sifnoc: Good point! I updated at [64bca73](https://github.com/zkmopro/mopro/pull/321/commits/64bca730550711c6ecf90d221b4a9b30ae586e2f)

PR #349, opened 2025-03-04
docs: benchmark viz

0xalizk:
Benchmarks are [plotted](https://github.com/0xalizk/mopro-benchmarks/tree/main) while the raw data are tucked under a collapsible section


vivianjeng:
I think it would make more sense if we can group circuits instead of proving libs (X-axis)
so the semaphore team can refer to the same group that which lib performs better
it doesn't make sense if I am the semaphore team, and I have to search each group and check which one performs the best

0xalizk:
I added both views tabulated: 

Screenshot 2025-03-10 at 4 31 59 PM

Screenshot 2025-03-10 at 4 32 07 PM

KimiWu123:
@0xalizk , what do you think about not using a log scale? It can highlight the dramatic difference, similar to what SP1 did, https://blog.succinct.xyz/sp1-benchmarks-8-6-24/

0xalizk:
@KimiWu123 Without log-scaling some bars would be too [tiny](https://github.com/0xalizk/mopro-benchmarks/tree/main/scale/nolog/README.md) to see. Here is how different scaling looks like:

- [log10](https://github.com/0xalizk/mopro-benchmarks/blob/main/scale/log10/README.md) (current)
- [log6](https://github.com/0xalizk/mopro-benchmarks/tree/main/scale/log6/README.md)
- [log3](https://github.com/0xalizk/mopro-benchmarks/tree/main/scale/log3/README.md)
- [log2](https://github.com/0xalizk/mopro-benchmarks/tree/main/scale/log2/README.md)
- [linear (no log)](https://github.com/0xalizk/mopro-benchmarks/tree/main/scale/nolog/README.md)


Not that by enforcing the display of certain ticks, by manually specifiying them, can shrink or expand the bars; compare [auto ticks](https://github.com/0xalizk/mopro-benchmarks/blob/main/scale/log10/plots/circom.android.wit.gen_by.bench.png) vs [manual ticks](https://github.com/0xalizk/mopro-benchmarks/blob/main/plots/circom.android.wit.gen_by.bench.png) of the same data

I modified the code to make both the scale (log or not) and the base (if log) [parametizable](https://github.com/0xalizk/mopro-benchmarks/tree/main/data), both optional with defaults to `linear` and `base=10` if it's log.

0xalizk:
Switched to linear scale

PR #351, opened 2025-03-06
Witnesscalc rapidsnark e2e 2

vivianjeng:
- related to #339 #310 
- The PR is to set `ProofLib` in `generate_circom_proof` and `verify_circom_proof`
- setup the environment for building `witnesscalc` and `rapidsnark`, but the tests are unstable: https://github.com/zkmopro/mopro/actions/runs/13645955562/job/38145379077
   I think we can fix this in the future (#352) and make the `ProofLib` be merged first
- setup the environment for android emulator tests


circom-prover/Cargo.toml
```diff
@@ -81,10 +81,10 @@ ark-poly = { version = "=0.4.1", default-features = false, features = [
 rand = { version = "0.8", features = ["std"] }
 
 # witnesscalc-adapter
-witnesscalc-adapter = { version = "0.1.0", optional = true }
+witnesscalc-adapter = { git = "https://github.com/zkmopro/witnesscalc_adapter.git", branch = "dylib", optional = true }
```
KimiWu123, line 84: Curious what the difference betwee `0.1.0` and `dylib` branch?

docs/docs/adapters/circom.md
```diff
@@ -130,12 +130,12 @@ The Circom adapter exposes the following functions to be used in the iOS project
 ```swift
 // Generate a proof for a given circuit zkey, as well as the circuit inputs
 // Make sure that the name of the zkey file matches the one you set in the `set_circom_circuits!` macro
-generateCircomProof(zkeyPath: zkeyPath, circuitInputs: inputs) -> GenerateProofResult
+generateCircomProof(zkeyPath: zkeyPath, circuitInputs: input, proofLib: proofLib) -> GenerateProofResult
```
KimiWu123, line 133: It's minor, but I would suggest to let `proofLib` be the first argument.

docs/docs/adapters/circom.md
```diff
@@ -130,12 +130,12 @@ The Circom adapter exposes the following functions to be used in the iOS project
...
 
 // Verify a proof for a given circuit zkey
 // This works for arbitrary circuits, as long as the zkey file is valid
 verifyCircomProof(
-    zkeyPath: zkeyPath, proof: generateProofResult.proof, publicInput: generateProofResult.inputs) -> Bool
+    zkeyPath: zkeyPath, proof: generateProofResult.proof, publicInput: generateProofResult.inputs, proofLib: proofLib) -> Bool
```
KimiWu123, line 138: the same as above.

KimiWu123 (approved):
LGTM! Only some minor suggestions and questions.

sifnoc:
One discussion point concerns the handling of `ProofLib` as an uniffi-exposed type. 

In my last PR, I applied a procedural macro instead of UDL. I originally assumed that the UDL file served as a customizable point for developers to when they modify functions -`prove` and `verify` (e.g., `generate_circom_proof`). Please correct me if I’m mistaken.

Assuming my understanding is correct, if `ProofLib` is intended to be customizable in the future, it should remain uniffi-exposed, as shown in the current PR. Otherwise, it might be better to control the behavior of `generate_circom_proof` using feature flags (similar to how `circom-prover` used them - `arkworks` and `rapidsnark`).

circom-prover/Cargo.toml
vivianjeng: `uniffi` needs to link to dylib https://github.com/zkmopro/witnesscalc_adapter/pull/12/files
I will publish a new version of `witnesscalc-adapter` if the e2e tests pass

docs/docs/adapters/circom.md
vivianjeng: I even want to make it an optional input
like in typescript, you can set an input to be default value
```ts
generateCircomProof (
   zkeyPath: string, 
   circuitInputs: string, 
   proofLib: ProofLib = ProofLib.arkworks
)
```
so users can use
either 
`generateCircomProof(zkeyPath, inputs)`
or 
`generateCircomProof(zkeyPath, inputs, ProofLib.rapidsnark)`

not sure if rust can do something like this
so we don't have to change the previous API

vivianjeng:
@sifnoc yes we have considered to use feature flags to control proof lib
but if we want to build an example benchmark app
It limits the usage
we are not able to enable both proof libs with feature flags
so I prefer to keep it like this (in current stage)

docs/docs/adapters/circom.md
KimiWu123: Unfortunately, Rust doesn't support default function agruments. 😢

test-e2e/Config.toml
sifnoc, line 1: Is this leftover?

sifnoc (approved):
LGTM!

test-e2e/Config.toml
vivianjeng: no I used it for `mopro build` :)

PR #371, opened 2025-03-16
chore: fixed typo in panic message

mdqst:
I found a tiny typo in the following line of code:

```rust
                    panic!("Invalid input, lager than MODULUS")
```

The word **"lager"** should be **"larger"**. Here's the corrected version:

```rust
                    panic!("Invalid input, larger than MODULUS")
```

While "lager" is great for a cold drink, it's not quite what we need here.
The message is clearly comparing numbers, so "larger" is the way to go.
Let’s save the beer for later and make sure the logic is on point!


vivianjeng:
serialization.rs will be removed in #375

Issue #368, opened 2025-03-14
Support circom proof convert for both BN254 and BLS12-381


vivianjeng:
# Problem
Now we implemented `toEthereumProof` and `fromEthereumProof`
https://github.com/zkmopro/mopro/blob/da02fc4b45957b9c834799a75d6f20914c7a3bff/mopro-ffi/src/circom/ethereum.rs#L27-L28
But it only supports `Bn254` for now
We can add support for `BLS12-381`

Discussion: https://github.com/zkmopro/mopro/pull/367#discussion_r1993746303
Discussion: https://github.com/zkmopro/mopro/pull/361#discussion_r1990727629

### Solution 1 (best)
Convert the proofs in `circom-prover` with the types: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/snarkjs/index.d.cts
so that the struct becomes
```rs
pub struct CircomProof {
    pub a: G1, // each element is BigUint type
    pub b: G2,
    pub c: G1,
    pub protocol: String,
    pub curve: String,
}

// circom-prover
pub struct GenerateProofResult {
    pub proof: CircomProof,
    pub inputs: Vec<BigUint>,
}

// mopro-ffi
pub struct GenerateProofResult {
    pub proof: CircomProof, // convert all BigUint to String
    pub inputs: Vec<String>, // convert all BigUint to String
}
```

the API becomes
```rs
// circom-prover
pub fn prove(
        proof_lib: ProofLib,
        wit_fn: WitnessFn,
        json_input_str: String,
        zkey_path: String,
    ) -> Result<CircomProof> {

pub fn verify(
        proof_lib: ProofLib,
        proof: CircomProof // merge proof and public_signals
        zkey_path: String,
    ) -> Result<bool> {
```

We have handled all BN254 and BLS12-381 convert in https://github.com/zkmopro/mopro/blob/main/circom-prover/src/prover/ethereum.rs
Then it should support these 2 curves to convert to `CircomProof`
Just need to update the `mopro-ffi` and `test-e2e` types

### Solution 2 (short term, not recommended)
add support for `toEthereumProof` with
https://github.com/zkmopro/mopro/blob/da02fc4b45957b9c834799a75d6f20914c7a3bff/mopro-ffi/src/circom/ethereum.rs#L28-L29
deserialize/serialize BLS12-381

## Acceptance criteria
Support for both BN254 and BLS12-381 that the proof needs to be
1. readable (like `toEthereumProof`)
[...]

PR #374, opened 2025-03-24
feat/support BLS12-381

KimiWu123:
part of #368 

Rapidsnark doesn't support BLS12-381 so this PR only adding `BLS12-381` support on `etherum_input/proof`

### DIFF
--- a/circom-prover/src/prover/rapidsnark.rs
+++ b/circom-prover/src/prover/rapidsnark.rs
@@ -1,24 +1,17 @@
+use super::circom::{CURVE_BLS12_381, CURVE_BN254};
+use super::{ark_circom::read_proving_key, serialization::SerializableInputs, PublicInputs};
+use crate::prover::circom::PROTOCOL_GROTH16;
 use crate::CircomProof;
-use anyhow::Result;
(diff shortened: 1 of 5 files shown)


circom-prover/src/prover/rapidsnark.rs
```diff
@@ -43,7 +36,6 @@ pub fn generate_circom_proof(
     let c = <Bn254 as Pairing>::G1Affine::new_unchecked(c_x, c_y);
     let b1_x = Fq::from_str(proof_json["pi_b"][0][0].as_str().unwrap()).unwrap();
     let b1_y = Fq::from_str(proof_json["pi_b"][0][1].as_str().unwrap()).unwrap();
-    // let b1 = <Bn254 as Pairing>::G2Affine::new_unchecked(b1_x, b1_y);
```
vivianjeng, line 46: this can be removed?

vivianjeng (approved):
LGTM 
But I will remove `to_ethereum_proof`, `from_ethereum_proof` like mentioned in #368 
and because the output of `CircomProof` and `ProofCalldata` is almost the same

PR #379, opened 2025-03-26
fix: fix circom-prover, update readme

vivianjeng:
- publish circom-prover 0.1.1

### DIFF
--- a/circom-prover/src/witness.rs
+++ b/circom-prover/src/witness.rs
@@ -46,8 +46,9 @@ pub fn generate_witness(witness_fn: WitnessFn, json_input_str: String) -> JoinHa
                 wit_fn(bigint_inputs)
             }
             #[allow(unreachable_patterns)]
-            _ => Vec::new(),
+            _ => panic!("Unsupported witness function"),
         };
+        #[allow(unreachable_code)]
         witness
             .into_iter()
             .map(|w| w.to_biguint().unwrap())


KimiWu123:
One more place we need to modify,
https://github.com/zkmopro/mopro/blob/3aebdc2e485016bf9a7066eb6aabf2db9421cc4e/mopro-ffi/src/lib.rs#L105
and 
https://github.com/zkmopro/mopro/blob/3aebdc2e485016bf9a7066eb6aabf2db9421cc4e/mopro-ffi/src/lib.rs#L112

`Option` is designed for `ProofCalldata`. Since we removed `ProofCalldata`, `z` of `G1` and `G2` should always have value.

circom-prover/src/witness.rs
```diff
@@ -1,30 +1,30 @@
...
 /// Witness function signature for rust_witness (inputs) -> witness
-#[cfg(feature = "rustwitness")]
 type RustWitnessWtnsFn = fn(HashMap<String, Vec<BigInt>>) -> Vec<BigInt>;
 /// Witness function signature for witnesscalc_adapter (inputs) -> witness
-#[cfg(feature = "witnesscalc")]
 type WitnesscalcWtnsFn = fn(&str) -> anyhow::Result<Vec<u8>>;
 
 #[derive(Debug, Clone, Copy)]
 pub enum WitnessFn {
-    #[cfg(feature = "witnesscalc")]
```
KimiWu123, line 16: If we removed this feature config, it could cause an unexpected error. `generate_witness` might return an empty array in L49 which causes proving error. And it's a runtime error, harder to troublshooting. Is it your intention?

KimiWu123 (commented):
Looks nice, left one question.

KimiWu123 (approved):

PR #383, opened 2025-03-28
fix: fix halo2 interface

vivianjeng:


mopro-ffi/src/lib.rs
```diff
@@ -64,8 +64,8 @@ macro_rules! halo2_app {
         fn verify_halo2_proof(
             srs_path: String,
             vk_path: String,
-            proof_data: Vec<u8>,
-            public_inputs: Vec<u8>,
+            proof: Vec<u8>,
+            public_input: Vec<u8>,
```
KimiWu123, line 68: curious when we use "input`s`" and when we use "input". Above `circuit_inputs` is plural but here is singular.

vivianjeng: just a quick fix for matching current halo2 interface
https://github.com/zkmopro/mopro/blob/c092d0daa08e684bbd61782f8148ddabbcb86c55/mopro-ffi/src/halo2/mod.rs#L10-L14
https://github.com/zkmopro/mopro/blob/c092d0daa08e684bbd61782f8148ddabbcb86c55/mopro-ffi/src/halo2/mod.rs#L27-L32
or the iOS will throw error when building
we could change to a better name in the future (but not urgent)

vivianjeng: if there will be another fix
we should also change
https://github.com/zkmopro/mopro/blob/c092d0daa08e684bbd61782f8148ddabbcb86c55/test-e2e/ios/mopro-test/ContentView.swift#L154
https://github.com/zkmopro/mopro/blob/c092d0daa08e684bbd61782f8148ddabbcb86c55/test-e2e/ios/mopro-test/ContentView.swift#L185-L186
and also CLI templates

PR #389, opened 2025-04-02
fix: fix cmake version

vivianjeng:
https://github.com/turbolent/w2c2/pull/115


vivianjeng:
w2c already fixed the cmake version
https://github.com/turbolent/w2c2/pull/115

PR #397, opened 2025-04-16
Fix CSS properties

sky-coderay:
### Description:

-Added missing “px” unit for width
-Added missing semicolon after the height property


vivianjeng:
It's a minor change so we won't merge this
thank you for the contribution

PR #404, opened 2025-04-28
feat/customize the lib name of mopro bindings

### LINKED ISSUE #387: Customized lib name of the mopro bindings
# Problem
The UniFFI provides different library name of the bindings
https://mozilla.github.io/uniffi-rs/0.29/tutorial/Prerequisites.html
We might also take advantage of this

## Details
The name of the bindings should be changed to customized lib name, e.g. `libmath` in the UniFFI example
### iOS
https://github.com/zkmopro/mopro/blob/f216f0b1e945e5629b8d2c5d4c684c763937a9f4/mopro-ffi/src/app_config/ios.rs#L56
https://github.com/zkmopro/mopro/blob/f216f0b1e945e5629b8d2c5d4c684c763937a9f4/mopro-ffi/src/app_config/ios.rs#L81
https://github.com/zkmopro/mopro/blob/f216f0b1e945e5629b8d2c5d4c684c763937a9f4/mopro-ffi/src/app_config/ios.rs#L104

### Android
https://github.com/zkmopro/mopro/blob/f216f0b1e945e5629b8d2c5d4c684c763937a9f4/mopro-ffi/src/app_config/android.rs#L95
https://github.com/zkmopro/mopro/blob/f216f0b1e945e5629b8d2c5d4c684c763937a9f4/mopro-ffi/src/app_config/android.rs#L100

## Acceptance criteria
Remove these `mopro_bindings`

## Next steps
Might need to change the CLI or check if it works for different names of libraries

KimiWu123:
To close #387

### DIFF
--- a/cli/src/build.rs
+++ b/cli/src/build.rs
@@ -60,7 +60,7 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
 
     // Platform selection
     let mut platform: PlatformSelector = match arg_platforms {
-        None => PlatformSelector::select(&config),
+        None => PlatformSelector::select(&mut config),
         Some(p) => {
             let valid_platforms: Vec<String> = p
                 .iter()
@@ -72,7 +72,7 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
                 style::print_yellow(
                     "No platforms selected. Please select at least one platform.".to_string(),
                 );
-                PlatformSelector::select(&config)
+                PlatformSelector::select(&mut config)
             } else {
                 if valid_platforms.len() != p.len() {
                     style::print_yellow(
@@ -147,7 +147,8 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
     }
 
     // Architecture selection for iOS or Android
-    let selected_architectures = platform.select_archs();
+    let selected_architectures = platform.select_archs(&mut config);
+    write_config(&config_path, &config)?;
 
     // Noir only supports `aarch64-apple-ios` and `aarch64-linux-android`
     if config.adapter_contains(Adapter::Noir) {
@@ -194,10 +195,7 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
             command.status()?
         };
 
-        // Add only successfully compiled platforms to the config.
-        if status.success() {
-            config.target_platforms.insert(platform_str.into());
-        } else {
+        if !status.success() {
             // Return a custom error if the command fails
             return Err(anyhow::anyhow!(
                 "Output with status code {}",
@@ -205,9 +203,6 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
             ));
         }
     }
-
-    // Save the updated config to the file
-    write_config(&config_path, &config)?;
 
     print_binding_message(&platform.platforms)?;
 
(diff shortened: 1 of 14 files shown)


vivianjeng (commented):
I am thinking about the solution like this
https://github.com/mozilla/uniffi-rs/blob/182d9692eeebd5c38897cb42f47c5560519207ef/uniffi_bindgen/src/library_mode.rs#L84

it will automatically find the `lib_name`
so we don't have to set the lib name here
https://github.com/zkmopro/mopro/blob/3a3b3eba7805395effb250c0cdcc7501b7b079dc/test-e2e/Cargo.toml#L7

and the bindings will also be `MoproiOSBindings` and `MoproAndroidBindings`
but if we have more than one bindings from apps
e.g. Semaphore use mopro to build the library with a `MoproiOSBindings` and Anon Aadhaar has another library with a `MoproiOSBindings`
then it causes the conflict in the XCode

and we need to figure out if we can change the name like
https://github.com/zkonduit/ezkl-swift-package/blob/ff7151928d75f108e91bfad3d5e5df8faf13f70d/Sources/EzklCoreBindings/EzklCore.swift#L11
Now all the bindings are `import moproFFI`
but it might be other libraries like I mentioned above

but the above are not considered in https://github.com/zkmopro/mopro/issues/387
I think we can make the scope in #387 bigger
or do you think we should open other issues to solve this?

KimiWu123:
What do you think we keep `name` for flexibility? Users can modify the name as they want. I can submit another commit to  make the lib name be the same as the project name (when users call `mopro init`)

mopro-ffi/src/app_config/mod.rs
```diff
@@ -55,3 +56,28 @@ pub fn install_arch(arch: String) {
         .wait()
         .unwrap_or_else(|_| panic!("Failed to install target architecture {}", arch));
 }
+
```
KimiWu123, line 59: If we don't specify the lib name, the output lib will be named as the package name. The only difference is '-' is replace with '_'. So, here I just fetch the package name and replace '-'. In addition, still keep the flexibility to let users can specify the lib name by setting `name` under `[lib]` section in the toml file.

cli/src/template/init/Cargo.toml
```diff
@@ -7,7 +7,6 @@ edition = "2021"
 # This library name should not be changed
 [lib]
 crate-type = ["lib", "cdylib", "staticlib"]
-name = "mopro_bindings"
```
KimiWu123, line 10: removed the lib `name` and the output lib name will be the package name.

cli/src/build.rs
```diff
@@ -147,7 +147,8 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
     }
 
     // Architecture selection for iOS or Android
-    let selected_architectures = platform.select_archs();
+    let selected_architectures = platform.select_archs(&mut config);
+    write_config(&config_path, &config)?;
```
KimiWu123, line 151: Write the config file before build to avoid a failed build causes users have to select again.

vivianjeng (approved):
LGTM!
It works well for me
but the folder structure is like
```
├── MoproAndroidBindings
│   ├── jniLibs
│   │   └── arm64-v8a
│   └── uniffi
│       └── mopro
├── MoproiOSBindings
│   ├── mopro.swift
│   └── MoproBindings.xcframework
│       ├── Info.plist
│       └── ios-arm64
```

There are still many places we hardcoded as `Mopro` and we can figure out how to improve this
I will draft another issue to describe the situation

PR #417, opened 2025-05-08
refactor: change ios and android fields in Config to Option<HashSet<S…

### LINKED ISSUE #415: `Config.toml` is not compatible with order version
# Problem
A new config is written to `Config.toml`, which causes the error like
```sh
> mopro build
Failed to build project: TOML parse error at line 1, column 1
  |
1 | target_adapters = []
  | ^^^^^^^^^^^^^^^^^^^^
missing field `ios`
```
Therefore if there are projects built before the PR #404 
then it will cause error

## Details

We can let the `ios` not required for `mopro build`

## Acceptance criteria
It should compatible with older version of mopro project

vivianjeng:
…tring>> for better handling of empty states
fix #415 

If there is no `ios` or `android` in `Config.toml`, then select all targets
If new targets are selected, then update the `Config.toml`

### DIFF
--- a/cli/src/build.rs
+++ b/cli/src/build.rs
@@ -6,6 +6,7 @@ use dialoguer::Confirm;
 use dialoguer::Select;
 use include_dir::include_dir;
 use include_dir::Dir;
+use std::collections::HashSet;
 use std::env;
 
 use crate::config::read_config;
@@ -148,6 +149,16 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
 
     // Architecture selection for iOS or Android
     let selected_architectures = platform.select_archs(&mut config);
+    let mut ios_hash = HashSet::new();
+    let mut android_hash = HashSet::new();
+    for arch in selected_architectures.get("ios").unwrap() {
+        ios_hash.insert(arch.clone());
+    }
+    for arch in selected_architectures.get("android").unwrap() {
+        android_hash.insert(arch.clone());
+    }
+    config.ios = Some(ios_hash);
+    config.android = Some(android_hash);
     write_config(&config_path, &config)?;
 
     // Noir only supports `aarch64-apple-ios` and `aarch64-linux-android`
--- a/cli/src/build.rs
+++ b/cli/src/build.rs
@@ -6,7 +6,6 @@ use dialoguer::Confirm;
 use dialoguer::Select;
 use include_dir::include_dir;
 use include_dir::Dir;
-use std::collections::HashSet;
 use std::env;
 
 use crate::config::read_config;
@@ -149,16 +148,6 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
 
     // Architecture selection for iOS or Android
     let selected_architectures = platform.select_archs(&mut config);
-    let mut ios_hash = HashSet::new();
-    let mut android_hash = HashSet::new();
-    for arch in selected_architectures.get("ios").unwrap() {
-        ios_hash.insert(arch.clone());
-    }
-    for arch in selected_architectures.get("android").unwrap() {
-        android_hash.insert(arch.clone());
-    }
-    config.ios = Some(ios_hash);
-    config.android = Some(android_hash);
     write_config(&config_path, &config)?;
 
     // Noir only supports `aarch64-apple-ios` and `aarch64-linux-android`
--- a/cli/src/utils.rs
+++ b/cli/src/utils.rs
@@ -100,9 +100,7 @@ impl PlatformSelector {
                     .iter()
                     .map(|&i| {
                         let arch = IosArch::from_idx(i).as_str().to_string();
-                        if let Some(ref mut ios) = config.ios {
-                            ios.insert(arch.clone());
-                        }
+                        config.ios.as_mut().unwrap().insert(arch.clone());
                         arch
                     })
                     .collect::<Vec<String>>();
@@ -129,9 +127,7 @@ impl PlatformSelector {
                     .iter()
                     .map(|&i| {
                         let arch = AndroidArch::from_idx(i).as_str().to_string();
-                        if let Some(ref mut android) = config.android {
-                            android.insert(arch.clone());
-                        }
+                        config.android.as_mut().unwrap().insert(arch.clone());
                         arch
                     })
                     .collect::<Vec<String>>();
--- a/cli/src/utils.rs
+++ b/cli/src/utils.rs
@@ -33,12 +33,12 @@ impl PlatformSelector {
     pub fn select(config: &mut Config) -> Self {
         let platforms = Platform::all_strings();
         // defaults based on previous selections or all platforms
-        let defaults: Vec<bool> = if config.target_platforms.is_empty() {
-            vec![true; platforms.len()]
+        let defaults: Vec<bool> = if config.target_platforms.is_none() {
+            vec![false; platforms.len()]
         } else {
             platforms
                 .iter()
-                .map(|&platform| config.target_platforms.contains(platform))
+                .map(|&platform| config.target_platforms.as_ref().unwrap().contains(platform))
                 .collect()
         };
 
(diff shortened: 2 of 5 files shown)


cli/src/utils.rs
```diff
@@ -80,24 +80,28 @@ impl PlatformSelector {
                 // defaults based on previous selections if previous selections are not empty
                 // otherwise, selects all
                 let all_ios_archs = IosArch::all_strings();
-                let defaults: Vec<bool> = if config.ios.is_empty() {
+                let defaults: Vec<bool> = if config.ios.is_none() {
```
KimiWu123, line 83: We still need `is_empty` here. The hashset object exsits (aka. `is_some()`) still can be an empty one.

cli/src/build.rs
```diff
@@ -148,6 +149,16 @@ pub fn build_project(arg_mode: &Option<String>, arg_platforms: &Option<Vec<Strin
...
     let selected_architectures = platform.select_archs(&mut config);
+    let mut ios_hash = HashSet::new();
+    let mut android_hash = HashSet::new();
+    for arch in selected_architectures.get("ios").unwrap() {
+        ios_hash.insert(arch.clone());
+    }
+    for arch in selected_architectures.get("android").unwrap() {
+        android_hash.insert(arch.clone());
+    }
+    config.ios = Some(ios_hash);
```
KimiWu123, line 160: Wondering why updating `ios` and `android` here instead of updating inside `select_archs` as original implementation?

I think the code is clearer if we move these update logic inside `select_archs()` and I did the same for `select()` to update `target_platforms`. If you prefer to move them outside, then I would suggest
1. do the same thing to `select()` (move `target_platforms` out)
2. try to remvoe `mut` for `config` when calling `select_archs()`, `mut` should not be necessary

cli/src/utils.rs
```diff
@@ -80,24 +80,28 @@ impl PlatformSelector {
...
                     all_ios_archs
                         .iter()
-                        .map(|&acrch| config.ios.contains(acrch))
+                        .map(|&acrch| config.ios.as_ref().unwrap().contains(acrch))
                         .collect()
                 };
 
                 // clear previous selections before update
-                config.ios.clear();
+                if let Some(ref mut ios) = config.ios {
```
KimiWu123, line 93: We can just new a `ios` here, assigning from `config.ios` and then clear it is not necessary.

cli/src/utils.rs
```diff
@@ -80,24 +80,28 @@ impl PlatformSelector {
...
+                    ios.clear();
+                }
 
                 let sel = Self::select_multi_archs(p.as_str(), &all_ios_archs, defaults);
                 let sel_str = sel
                     .iter()
                     .map(|&i| {
                         let arch = IosArch::from_idx(i).as_str().to_string();
-                        config.ios.insert(arch.clone());
+                        if let Some(ref mut ios) = config.ios {
```
KimiWu123, line 102: the same here, we don't need `config.ios`, just update `ios` directly is enough.

cli/src/utils.rs
KimiWu123: the same as `android` below

cli/src/utils.rs
KimiWu123: the same as `android` below

cli/src/utils.rs
KimiWu123: the same as `android` below

cli/src/build.rs
vivianjeng: right it doesn't have to be written again
I removed them all [029aaec](https://github.com/zkmopro/mopro/pull/417/commits/029aaec3858a5897fbbc747d4efea1752807973e)

cli/src/utils.rs
vivianjeng: I decided not to let it default select all
so when it's empty or it's len = 0
none of them will be selected
[029aaec](https://github.com/zkmopro/mopro/pull/417/commits/029aaec3858a5897fbbc747d4efea1752807973e)

cli/src/utils.rs
vivianjeng: [029aaec](https://github.com/zkmopro/mopro/pull/417/commits/029aaec3858a5897fbbc747d4efea1752807973e)

cli/src/utils.rs
vivianjeng: [029aaec](https://github.com/zkmopro/mopro/pull/417/commits/029aaec3858a5897fbbc747d4efea1752807973e)

cli/src/utils.rs
KimiWu123: cuious why you decided it? I think selecting all in default is nice. BTW, `is_none` and `is_empty` is different.

cli/src/utils.rs
KimiWu123: `config.ios` was re-created above (`config.ios = Some(HashSet::new());`) so we don't need check if it's `Some`.

KimiWu123:
I have a different idea on this issue,  it's a backward compatible issue basically. I think we can modify `read_config` to prevent from this issue and we don't need to add `Some` on `android` and `ios`. it seems easier and cleaner for me.

vivianjeng:
@KimiWu123 
I think even if we don't use `Option`, we still need another structure to serialize/deserialize the toml
e.g.
```rust
// Storing user selections while iterating with mopro cli
#[derive(Debug, serde::Deserialize, serde::Serialize)]
pub struct TargetConfig {
    pub(crate) target_adapters: HashSet<String>,
    pub(crate) target_platforms: HashSet<String>,
}
```

if we want to reuse the `Config.toml` and save to `Config`
we still need to read like `let current_config: TargetConfig = toml::from_str(&contents)?;`

and in toml documentation https://docs.rs/toml/latest/toml/
the `Option` is suggested if a field is optional

so I think make the field optional is better
e.g.

```rust
// Storing user selections while iterating with mopro cli
#[derive(Debug, serde::Deserialize, serde::Serialize)]
pub struct Config {
    pub(crate) target_adapters: HashSet<String>,
    pub(crate) target_platforms: HashSet<String>,
    pub(crate) ios: Option<HashSet<String>>,
    pub(crate) android: Option<HashSet<String>>,
}
```

KimiWu123:
> @KimiWu123 I think even if we don't use `Option`, we still need another structure to serialize/deserialize the toml e.g.
> 
> ```rust
> // Storing user selections while iterating with mopro cli
> #[derive(Debug, serde::Deserialize, serde::Serialize)]
> pub struct TargetConfig {
>     pub(crate) target_adapters: HashSet<String>,
>     pub(crate) target_platforms: HashSet<String>,
> }
> ```
> 
> if we want to reuse the `Config.toml` and save to `Config` we still need to read like `let current_config: TargetConfig = toml::from_str(&contents)?;` 

I didn't get it. We had two functions `read_config` and `write_config` already to read from  and write to the `config.toml`. I don't understand why we need another struct to store `target_adapters` and `target_platforms` only. My implementation was trying to collect all the configs in the same file (`config.toml`).

> and in toml documentation https://docs.rs/toml/latest/toml/ the `Option` is suggested if a field is optional
> 

I don't get why `ios` and `android` is optional. By introducing these two fields, users don't need to select archs again if they fails in any steps   
I prefer making it easier by handling this kind of backward compatible logic in `read_config` and `write_config`, and we can make sure the inputs `config` is well-handled when we're using it. That being said, I don't have strong opions on adding `Option` or not. The thing I want to mention is  if we add `Option` on `ios` and `android`, I think we need to add on `target_adapters` and `target_platforms` as well. I don't see any difference among these 4 variables. Beside, as I mentioned above, `select()` should be modify as well (not modifying `config` object as what you did on `select_archs()`, returning the result).

I can understand this is a small PR, but I want to highlight here is that we still should keep our code consistent.

cli/src/utils.rs
vivianjeng: 1. it takes time to build
2. some of them are not commonly used nowadays (e.g. `x86_64-apple-ios`, all the other android targets except for `aarch64-linux-android`)
3. some proving systems don't support all targets
    e.g. Noir, `rapidsnark`/`witnesscalc`

cli/src/utils.rs
vivianjeng: [76e1642](https://github.com/zkmopro/mopro/pull/417/commits/76e164275fd080011eca11d337734fa621d7c357)

KimiWu123 (approved):
LGTM! Feel free to merge it after the testings pass.

Issue #414, opened 2025-05-05
[Noir] benchmark for different platforms


vivianjeng:
# Problem
Now the mopro supports Noir #353 #398 
Then we can provide Noir benchmark in the documentation and the Readme

## Details
<table>
  <tr>
    <th>Circuits</th>
    <th>iOS</th>
    <th>Android</th>
    <th>Laptop</th>
    <th>Website</th>
  </tr>
  <tr>
    <td>Keccak256</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
  </tr>
  <tr>
    <td>RSA</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
  </tr>
<tr>
    <td>zkemail</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
  </tr>
<tr>
    <td>anon aadhaar</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
  </tr>
<tr>
    <td>semaphore</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
  </tr>
</table>

- zkemail: https://github.com/zkemail/zkemail.nr
- anon aadhaar: https://github.com/anon-aadhaar/anon-aadhaar-noir
- semaphore: https://github.com/hashcloak/semaphore-noir (not sure if sempahore team has another repo)

## Acceptance criteria
Update the data here https://zkmopro.org/docs/performance and here https://github.com/zkmopro/mopro?tab=readme-ov-file#performance

moven0831:
## update

* On iOS, the benchmark runs using the “My Mac (Designed for iPad)” target.
* On Android, it runs on a Pixel 6 in release mode.
* On laptops, we measure wall-clock time with the `time bb prove` command.
* On the website, the benchmark runs in a browser-based (WASM) environment.


<table>
  <tr>
    <th>Circuits</th>
    <th>iOS</th>
    <th>Android</th>
    <th>Laptop</th>
    <th>Website</th>
  </tr>
  <tr>
    <td>Keccak256</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
    <td> ? s</td>
  </tr>
  <tr>
    <td>RSA</td>
    <td> ? s</td>
    <td> ? s</td>
[...]

PR #436, opened 2025-05-21
docs: noir benchmark

moven0831:
This PR introduces Noir benchmark results to the project, updating both the main `README.md` and the detailed performance documentation in `docs/docs/performance.md`.

- related issue: #414

### DIFF
--- a/docs/docs/performance.md
+++ b/docs/docs/performance.md
@@ -414,3 +414,28 @@ The below tests were run on a Macbook Pro M1 Pro (2021) as well as an iPhone 15 
 Note that the iPhone 15 Pro and Pixel 6 Pro crash when running the RSA circuit due to the large memory requirements. The circuit needs
 around 5GB of memory to run, while the iPhone 15 Pro and Pixel 6 Pro usually limits the application memory usage to 3GB.
 Note that the RSA circuit built on top of former version of Halo2. When it generates a proof on web, it crashes inside Halo2 module.
+
+## Noir
+
+We target [`nargo v1.0.0-beta.3`](https://noir-lang.org/docs/getting_started/quick_start#nargo) and [`bb v0.82.2`](https://noir-lang.org/docs/getting_started/quick_start#proving-backend-1), ensuring all related dependencies are aligned (e.g., [@noir-lang/noir_js@1.0.0-beta.3](https://www.npmjs.com/package/@noir-lang/noir_js/v/1.0.0-beta.3), [@aztec/bb.js@0.82.2](https://www.npmjs.com/package/@aztec/bb.js/v/0.82.2)). The benchmarks below were run on a MacBook Air M3 (2024) and a Pixel 6 (2021). We report the fastest result from multiple runs to reduce cold-start effects.
+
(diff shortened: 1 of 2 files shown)


docs/docs/performance.md
```diff
@@ -414,3 +414,28 @@ The below tests were run on a Macbook Pro M1 Pro (2021) as well as an iPhone 15
...
 around 5GB of memory to run, while the iPhone 15 Pro and Pixel 6 Pro usually limits the application memory usage to 3GB.
 Note that the RSA circuit built on top of former version of Halo2. When it generates a proof on web, it crashes inside Halo2 module.
+
+## Noir
+
+We target [`nargo v1.0.0-beta.3`](https://noir-lang.org/docs/getting_started/quick_start#nargo) and [`bb v0.82.2`](https://noir-lang.org/docs/getting_started/quick_start#proving-backend-1), ensuring all related dependencies are aligned (e.g., [@noir-lang/noir_js@1.0.0-beta.3](https://www.npmjs.com/package/@noir-lang/noir_js/v/1.0.0-beta.3), [@aztec/bb.js@0.82.2](https://www.npmjs.com/package/@aztec/bb.js/v/0.82.2)). The benchmarks below were run on a MacBook Air M3 (2024) and a Pixel 6 (2021). We report the fastest result from multiple runs to reduce cold-start effects.
+
+On iOS, the benchmark runs using the "My Mac (Designed for iPad)" target. On Android, it runs on a Pixel 6 in release mode. On bb CLI, we measure wall-clock time with the `time bb prove` command. On the website, the benchmark runs in a browser-based (WASM) environment.
```
KimiWu123, line 422: If the number comes from an emulator, I think we can't use it. All the benchmark result should come from real devices. Any idea, @vivianjeng ?

moven0831:
the table has updated with improvement factors @vivianjeng

docs/docs/performance.md
vivianjeng: good point
let me try on my iPhone

vivianjeng:
hey @moven0831 I found that the semaphore depth is 10
https://github.com/moven0831/mopro-example-app-semaphore-noir/blob/790f331dac2ab616c0885a721fdf450d9460a76b/circuit/src/main.nr#L9
but the current circom uses `depth = 32`
maybe we can align the depth?
it should support depth = 32??

docs/docs/performance.md
vivianjeng: I tried with iPhone device, and the result is quite similar
so I think we can keep this benchmark result

moven0831:
> hey @moven0831 I found that the semaphore depth is 10 https://github.com/moven0831/mopro-example-app-semaphore-noir/blob/790f331dac2ab616c0885a721fdf450d9460a76b/circuit/src/main.nr#L9 but the current circom uses `depth = 32` maybe we can align the depth? it should support depth = 32??

sure, it's been updated in the latest PR, please check.

interestingly, Noir's benchmark results are generally better than Circom's best performer (rapidsnark). however, for `semaphore-32`, it's slower. (e.g. for macOS, noir: 436 ms; rapidsnark: 112 ms)

PR #450, opened 2025-06-09
feat(cli): update bindings with `mopro update`

### LINKED ISSUE #269: `mopro update` command
# Problem
When a user go through
`mopro init`, `mopro build` and `mopro create`
he can get a template with bindings in the first `mopro build`
but when the user wants to 
1. update circuits
2. update UDL

He has to manually copy the bindings to the template
like https://github.com/zkmopro/react-native-app?tab=readme-ov-file#3-update-mopro-bindings

## Justification 
DX improvement 

## Details

We can provide a more advanced command like `mopro update`
It finds (all*) the mopro bindings' paths and replace them with the latest `MoproiOSBindings` and `MoproAndroidBindings`
( *If the user create more than one template)

## Acceptance criteria

When executing `mopro update`
The bindings in the templates are updated. (e.g. with new circuits or with new functions)

vivianjeng:
- close #269

### DIFF
--- a/cli/src/update.rs
+++ b/cli/src/update.rs
@@ -3,7 +3,9 @@ use std::fs;
 use std::path::Path;
 use walkdir::WalkDir;
 
-use crate::constants::Platform;
+use crate::constants::{
+    Platform, JNILIBS_DIR, MOPRO_KOTLIN_FILE, MOPRO_SWIFT_FILE, XCFRAMEWORK_NAME,
+};
 use crate::print::print_update_success_message;
 use crate::style::{print_gray_items, print_green_bold};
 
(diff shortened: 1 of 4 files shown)


cli/src/update.rs
```diff
@@ -0,0 +1,121 @@
...
+        let updated_kotlin_paths = update_file(
+            &android_bindings_dir
+                .join("uniffi")
+                .join("mopro")
+                .join("mopro.kt"),
+            "mopro.kt",
+        )?;
+        print_gray_items(updated_kotlin_paths);
+    }
+
```
KimiWu123, line 42:
```suggestion
// Update Wasm bindings
```

KimiWu123 (approved):
LGTM! Some nitpicks
1. I saw plain text `MoproBindings.xcframework` and `mopro.swift` defined in many places in `update.rs` and `create/flutter.rs` as well. Probably, we can try to use `Platform::Ios.binding_dir()` to replace all of them. `MoproWasmBindings` is the same.
2. Since we're planning to customize the binding's name, we could try not to hardcode the binding's name. Maybe passing the binding's name in `cli/src/main.rs`?
3. I also found that there are similar logics in the `fultter.rs`, [here](https://github.com/zkmopro/mopro/blob/main/cli/src/create/flutter.rs#L44-L46) and 
[here](https://github.com/zkmopro/mopro/blob/main/cli/src/create/flutter.rs#L57-L61). If possible, we should unify the update logic.

Issue #413, opened 2025-05-05
Customized bindings/xcframework name


vivianjeng:
# Problem
Now we finished the customized lib name: https://github.com/zkmopro/mopro/issues/387
but we found that there are still many un-customized folder/bindings
The bindings folder is like
```sh
├── MoproAndroidBindings
│   ├── jniLibs
│   │   └── arm64-v8a
│   └── uniffi
│       └── mopro
├── MoproiOSBindings
│   ├── mopro.swift
│   └── MoproBindings.xcframework
│       ├── Info.plist
│       └── ios-arm64
```

## Details
The most critical one is the `MoproBindings.xcframework`
If we want to integrate 2 bindings (might be different functions) and both include `MoproBindings.xcframework`, XCode will throw error when importing them though the lib name is different 

we should gradually remove the `Mopro` related names and make it works with current Mopro CLI and templates
- https://github.com/zkmopro/mopro/blob/d6936fe1183a6a02239b463851671feef863955c/mopro-ffi/src/app_config/ios.rs#L18
- https://github.com/zkmopro/mopro/blob/d6936fe1183a6a02239b463851671feef863955c/mopro-ffi/src/app_config/ios.rs#L30
- https://github.com/zkmopro/mopro/blob/d6936fe1183a6a02239b463851671feef863955c/mopro-ffi/src/app_config/ios.rs#L116
- https://github.com/zkmopro/mopro/blob/d6936fe1183a6a02239b463851671feef863955c/mopro-ffi/src/app_config/ios.rs#L200
- https://github.com/zkmopro/mopro/blob/d6936fe1183a6a02239b463851671feef863955c/mopro-ffi/src/app_config/android.rs#L22

May also refer to the uniffi documentation: https://mozilla.github.io/uniffi-rs/latest/kotlin/configuration.html

## Acceptance criteria

If we build 2 bindings with `mopro build`, the example project still work when importing them both
Follow
- https://zkmopro.org/docs/setup/ios-setup
- https://zkmopro.org/docs/setup/android-setup
- https://zkmopro.org/docs/setup/react-native-setup
- https://zkmopro.org/docs/setup/flutter-setup
- (https://zkmopro.org/docs/setup/web-wasm-setup) (will be nice)

vivianjeng:
https://github.com/zkmopro/mopro/pull/450#pullrequestreview-2910912300

vivianjeng:
[...]

PR #482, opened 2025-06-23
feat:  Enable multi-binding support on iOS and Android while preserving backward compatibility

ElusAegis:
This PR addresses **Issue #413**,  enabling multiple UniFFI/Mopro bindings per mobile app while preserving compatibility with existing examples and documentation. Instead of introducing fully customizable binding names (useful only for very few scenarios), this PR quietly resolves internal symbol conflicts between bindings, keeping high-level naming intact and ensuring existing samples remain functional.

Our goal is to preserve simplicity for most users, who typically will never need custom or multiple bindings, while providing straightforward instructions for those who do. We believe this strikes the right balance between functionality and ease of use.

**Changes made:**

- **UniFFI scaffolding:**  
  No longer hardcodes the name as `mopro` (`uniffi::setup_scaffolding!()` instead of `uniffi::setup_scaffolding!("mopro")`). This automatically creates unique symbols, preventing linker conflicts.

- **iOS bindings:**  
  Header files and `module.modulemap` are now placed inside a crate-specific folder within the xcframework (`Headers/<crate>/<crate>FFI.h` instead of `Headers/<crate>FFI.h`). This preserves compatibility, with added documentation (Step 5) explaining how to configure multiple bindings in Xcode.

- **Android bindings:**  
  No structural changes. Added documentation (Step 6) provides clear instructions for configuring multiple bindings for Android. The CLI's `mopro update` command was improved to update bindings gracefully, leaving other existing bindings intact.


vivianjeng (approved):
Thank you for the PR
It works well for me!
But we can have some improvements in the future
1. the library name could be e.g.`MoproBindings` rather than `mopro_bindings` that will be more consistent with swift and kotlin naming
2. It would be better to have some tutorial images that can be shown in the documentation. It might be a little bit hard to find the path in the XCode.
3. also some example usages can be included
    ```swift
    import MyLibraryA
    import MyLibraryB

    let resultA = MyLibraryA.moproUniffiHelloWorld()
    let resultB = MyLibraryB.moproUniffiHelloWorld()
    ```

    ```kotlin
    val resultA = uniffi.mylibrarya.moproUniffiHelloWorld()
    val resultB = uniffi.mylibraryb.moproUniffiHelloWorld()
    ```

PR #489, opened 2025-06-25
feat: add optional `uniffi` re-export via `mopro_ffi`

### LINKED ISSUE #356: re-export uniffi
# Problem
Now the test e2e needs to import both `mopro-ffi` and `uniffi` to make it generate bindings

## Details
We can simplify this by re-exporting `uniffi` in `mopro-ffi`
so the users can just use

```toml
[dependencies]
mopro-ffi = "0.1"

[build-dependencies]
mopro-ffi = "0.1"
```

instead of
```toml
[dependencies]
mopro-ffi = "0.1"
uniffi = "0.29"

[build-dependencies]
mopro-ffi = "0.1"
uniffi = { version = "0.29", features = ["build"]}
```

We can also make it optional
e.g.
user can choose to use the 2nd way if he wants to use specific version of uniffi
and use a feature flag `uniffi` to export the default `uniffi` version
or it becomes
```toml
[dependencies]
mopro-ffi = {version = "0.1", features = ["uniffi"]}

[build-dependencies]
mopro-ffi = {version = "0.1", features = ["uniffi"]}
```

## Acceptance criteria
If the user removes `uniffi` dependencies, `mopro-ffi` can still generate bindings

ElusAegis:
Closes #356.

# Problem  
Downstream crates had to depend on both `mopro_ffi` and `uniffi` in `Cargo.toml`.  
This duplicated dependencies, risked version mismatches, and cluttered project manifests.

# What changed  

1. **UniFFI re-export**  
   The new `uniffi` feature (not enabled by default) re-exports all UniFFI crates items:

```rust
   #[cfg(feature = "uniffi")]
   pub use uniffi::*;
```

2. **Crate-alias macro**  
    UniFFI’s macros reference `::uniffi::*` directly.     To satisfy them, we introduce `uniffi_setup!()`, which shadows the crate name:

```rust
   #[cfg(feature = "uniffi")]
   #[macro_export]
   macro_rules! uniffi_setup {
       () => {
           extern crate mopro_ffi as uniffi;
       };
   }
```

   The `app!` macro now expands this automatically; downstream code needs no changes.

3. **Testing support**  
   Added a companion feature `uniffi-test` that re-exports UniFFI’s test helpers.

# Benefits  
- **Simpler manifests** - only `mopro_ffi` is required.  
- **Version safety** - UniFFI’s version is locked to the one shipped with Mopro.  
- **Cleaner code** - UniFFI macros resolve transparently through the alias.  
- **Backwards compatible** - the feature is opt-in and does not affect existing setups.

# Next steps  
- Look for a cleaner way to re-export the the UniFFI without overshadowing `uniffi` with `mopro_ffi`.
- Consider whether to enable the `uniffi` feature by default, with a compatibility break.

### DIFF
--- a/cli/src/create/write_toml.rs
+++ b/cli/src/create/write_toml.rs
@@ -45,7 +45,6 @@ default = ["<FEATURES>"]
 mopro-wasm = { git = "https://github.com/zkmopro/mopro.git" }
 mopro-ffi = { git = "https://github.com/zkmopro/mopro.git" }
 rust-witness = "0.1"
-uniffi = { version = "=0.29.0" }
 num-bigint = "0.4.0"
 thiserror = "2.0.12"
 # HALO2_DEPENDENCIES
@@ -53,9 +52,7 @@ thiserror = "2.0.12"
 # NOIR_DEPENDENCIES
 
 [build-dependencies]
-mopro-ffi = { git = "https://github.com/zkmopro/mopro.git" }
 rust-witness = "0.1"
-uniffi = { version = "=0.29.0", features = ["build"] }
 
 # CIRCOM_DEPENDENCIES
     "#
--- a/cli/src/create/write_toml.rs
+++ b/cli/src/create/write_toml.rs
@@ -43,7 +43,7 @@ default = ["<FEATURES>"]
 
 [dependencies]
 mopro-wasm = { git = "https://github.com/zkmopro/mopro.git" }
-mopro-ffi = { git = "https://github.com/zkmopro/mopro.git" }
+mopro-ffi = { git = "https://github.com/zkmopro/mopro.git", features = ["uniffi"] }
 rust-witness = "0.1"
 num-bigint = "0.4.0"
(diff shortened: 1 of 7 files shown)


cli/src/create/write_toml.rs
```diff
@@ -43,19 +43,16 @@ default = ["<FEATURES>"]
 
 [dependencies]
 mopro-wasm = { git = "https://github.com/zkmopro/mopro.git" }
-mopro-ffi = { git = "https://github.com/zkmopro/mopro.git" }
+mopro-ffi = { git = "https://github.com/zkmopro/mopro.git", features = ["uniffi"] }
```
vivianjeng, line 46: what if the feature is not activated?
users can import `uniffi` by theirselves?

ElusAegis: Yes, exactly. If the feature is not enabled, then they should define the dependency on `uniffi` in the Cargo.toml just as before

PR #516, opened 2025-08-03
docs: the test badge and link to it have been replaced

operagxsasha:


vivianjeng:
it is not a necessary change

Issue #503, opened 2025-07-11
Templates: Noir on-chain verifier


vivianjeng:
# Problem
Now https://github.com/zkmopro/noir-rs supports noir proof but we haven't figured out how to verify the proof on-chain.
We will need to update the noir/bb version and build a on-chain templates for devs to build for on-chain apps.

## Details
- Update noir-rs version like zkpassport https://github.com/zkpassport/aztec-packages/tree/v1.0.0-nightly.20250701
- Check if the proof can be verified on-chain
- build a template for devs to connect wallet and send the proof on-chain
   - react-navie
   - flutter

## Acceptance criteria
Build the template in a standalone repo
and it can be shown in https://zkmopro.org/docs/projects

PR #535, opened 2025-08-19
Update to `noir-rs` 1.0.0 beta.8-3 with `bb` 1.0.0-nightly.20250723

moven0831:
related issue:
- #503

## Summary
- Update [`noir-rs`](https://github.com/zkmopro/noir-rs) version from 1.0.0 beta.3-2 to 1.0.0 beta.8-3
- Update Barretenberg Backend ([`bb`](https://github.com/zkmopro/aztec-packages)) from 0.82.2 to 1.0.0-nightly.20250723
  - Introduce low memory mode for writing vk and generating proof
  - Provide a simple interface for developers to generate and verify noir proof with choices `on-chain mode` and `low memory mode`
      - Generate Noir Proof
        ```rs
          fn generate_noir_proof(
              circuit_path: String,     // Path to your circuit file
              srs_path: Option<String>, // Optional setup file path
              inputs: Vec<String>,      // Your input values
              on_chain: bool,          // For blockchain use?
              vk: Vec<u8>,            // Verification key
              low_memory_mode: bool    // Use less RAM?
          ) -> Result<Vec<u8>, Error>  // Returns: proof bytes
          ```
      - Verify Noir Proof
          ```rs
          fn verify_noir_proof(
              circuit_path: String,    // Same circuit file
              proof: Vec<u8>,         // The proof to check
              on_chain: bool,         // Same blockchain setting
              vk: Vec<u8>,           // Same verification key
              low_memory_mode: bool   // Same memory setting
          ) -> Result<bool, Error>    // Returns: true if valid
          ```
      - Generate Noir Verification Key
          ```rs
          fn get_noir_verification_key(
              circuit_path: String,     // Path to circuit
              srs_path: Option<String>, // Optional setup file
              on_chain: bool,          // Blockchain compatibility
              low_memory_mode: bool    // Memory optimization
          ) -> Result<Vec<u8>, Error>  // Returns: verification key
          ```
    Since the verification key is tied to the generated Solidity Verifier, the interface took vk as default parameters.
  - for demo of on-chain verification and wallet integration, see https://github.com/moven0831/mopro-wallet-connect-noir


moven0831:
propose: wrap `generate_noir_proof`, `verify_noir_proof` with `on-chain` boolean so that we can abstract the logic for developers about choosing using Poseidon or keccak hash

e.g. 
```
generate_noir_proof
|
| -- on-chain: true --  `prove_ultra_keccak_honk`
|
| -- on-chain: false -- `prove_ultra_honk` (default: Poseidon)
```

moven0831:
proposal: make `vk` a default parameter in `generate_noir_proof` and `verify_noir_proof` for better flexibility

cli/src/template/init/src/lib.rs
```diff
@@ -20,7 +20,73 @@ fn mopro_uniffi_hello_world() -> String {
...
+        let valid = verify_noir_proof(
+            circuit_path,
+            proof,
+            true,   // on_chain (uses Keccak for Solidity compatibility)
+            vk,
+            false,  // low_memory_mode
+        ).unwrap();
+        assert!(valid);
+    }
+}
```
vivianjeng, line 89: this will be in `//NOIR_TEMPLATE`?
so when others select different adapter, it will not be shown

vivianjeng (commented):
In the future, we should let the e2e tests almost the same as the cli templates
(ideally download the cli templates and run e2e tests)

moven0831:
> In the future, we should let the e2e tests almost the same as the cli templates (ideally download the cli templates and run e2e tests)

make sense, the root cause of the previous CI failure on Android test is race condition of proof generation for all kinds of proving systems. Tests could fail according to the execution order. It has been fixed in e5420f5d287926afe0e18908bfd8defca14bf52d with proper thread sleep and state clean.

https://github.com/moven0831/mopro/blob/3a31a5718da5c2569280452c0191fa7ad14f28f3/test-e2e/android/app/src/androidTest/java/com/mopro/mopro_app/ExampleInstrumentedTest.kt#L19-L41

PR #548, opened 2025-08-27
Updated FFI naming convention

### LINKED ISSUE #500: FFI: best naming of customized bindings
# Problem
Now the bindings' library name is built by 
https://github.com/zkmopro/mopro/blob/a297ede5be0c84011359a0a631f5c0e7dac14219/mopro-ffi/src/app_config/mod.rs#L60

and it is like `mopro_bindings`
but in Swift and Kotlin the naming is like `MoproBindings`

## Details
let the naming be first letter upper case and remove the `_`
but also need to be careful if there is other place requires to be `mopro_bindings`

## Acceptance criteria
Let the import library be
`libMoproBindings.a`
`import MoproBindings` in swift 
`import uniffi.MoproBindings.*` in kotlin

PoulavBhowmick03:
Fixes #500 

WIP


moven0831:
- Hi @PoulavBhowmick03 , since there's a simpler way to resolve the issue and the codebase is outdated compared with the current main branch, we will close this PR for now and move the discussion to #554

Issue #536, opened 2025-08-19
Refactor: refactor of `mopro-ffi`, `cli` and `test-e2e` to reduce dependencies


vivianjeng:
# Problem
- Now `cli` depends on `mopro-ffi`. But when we need to publish a crate like https://crates.io/crates/mopro-cli, the `noir` dependencies are not published, crates doesn't allow a dependency with source from git.
   - we need to publish cli to allow users to download mopro-cli like `cargo install mopro-cli` without cloning the repo
   - `cli` depends on `mopro-ffi` is to reduce the building time (https://github.com/zkmopro/mopro/pull/486)
- `test-e2e` doesn't check if `cli` works well. it uses `cargo run --bin ios` instead of `mopro build --platform ios`. And we should test all the platforms with mopro cli e.g. `mopro build --platform react-native`

### `mopro-ffi`
- Remove dependencies and code for `circom`, `halo2` and `noir` just keep the `app_config` folder
   - and add different FFIs in the future (e.g. `wasm`, `react-native` and `dart`)

### `cli`
- Add templates for `circom`, `halo2` and `noir`
   - add dependencies like `noir = {git ="https://..."}` as template, not really depends on it
   - add file like https://github.com/zkmopro/mopro/blob/main/mopro-ffi/src/circom/mod.rs to process the circom adapter
 
### `test-e2e`
- test for `mopro-cli`
   - mopro init
      - matrix: ["circom", "halo2", "noir"]
   - mopro build
      - matrix: ["ios", "android", "wasm", "react-native", "flutter"]
   - mopro create
      - matrix: ["ios", "android", "web", "react-native", "flutter"]
- run e2e tests in the templates
   -  rust tests
   - ios tests on simulator
   - android tests on simulator
   - web tests with jest
   - react-native e2e tests
   - flutter e2e tests

## Acceptance criteria
- Publish `mopro-ffi` and `cli` without depending on `circom`, `halo2` and `noir` dependencies
- Run e2e tests with `mopro-cli`

PR #556, opened 2025-09-15
fix: correct and unify library naming conventions

### LINKED ISSUE #500: FFI: best naming of customized bindings

moven0831:
- patch for #554 and resolve #500

### DIFF
--- a/mopro-ffi/src/app_config/mod.rs
+++ b/mopro-ffi/src/app_config/mod.rs
@@ -106,6 +106,8 @@ pub fn install_arch(arch: String) {
         .unwrap_or_else(|_| panic!("Failed to install target architecture {arch}"));
 }
 
+/// Gets the project name from Cargo.toml in snake_case format (e.g., "mopro_bindings")
+/// This is used for uniffi-style identifiers and library file names
 pub fn project_name_from_toml(project_dir: &Path) -> anyhow::Result<String> {
     let cargo_toml_path = project_dir.join("Cargo.toml");
     let cargo_toml_content =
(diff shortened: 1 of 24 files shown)


mopro-ffi/src/app_config/mod.rs
```diff
@@ -148,6 +150,22 @@ pub fn project_name_from_toml(project_dir: &Path) -> anyhow::Result<String> {
...
+        .map(|(i, c)| {
+            if c.is_uppercase() && i > 0 {
+                format!("_{}", c.to_lowercase())
+            } else {
+                c.to_lowercase().to_string()
+            }
+        })
+        .collect::<String>()
+}
+
```
moven0831, line 168: @vivianjeng , we now have utility functions for both snake case (for uniffi) and pascal case (mobile lib) identifiers. This keeps `android.rs` and `ios.rs` simple

moven0831: I see, then I think we can keep the `project_name_from_toml` as it is to get the project name in snake case, just like the original one. Then, use `snake_to_pascal_case` to convert it to upper case and make it compatible with `android.rs` and `ios.rs`

moven0831:
Hey @vivianjeng , it's a valid point that we should keep flutter app and RN app able to run directly after git clone.

In ddc64ce2a32629d8e7d670cf5deb13c0c747f7c6, I've adopted the update logic so that it can preserve the changes using `import uniffi.mopro.*` statements in
- https://github.com/zkmopro/flutter-app/pull/16
- https://github.com/zkmopro/react-native-app/pull/18
but also able to be updated with customized Android package names like `import uniffi.MyAndroidPackage.*` if the template is downloaded with `mopro create`

Moreover, for the native Android templates in `cli/src/template/android`, we can keep using `{{PACKAGE_NAME}}` for substitution logic since it will always be handled with `copy_embedded_dir_with_substitution` in `cli/src/create/android.rs` during `mopro create`

moven0831:
Close this PR first since it can be largely improved after the FFI refactor in #492, #536, #540, and #561

PR #588, opened 2025-09-30
feat: Combined 3 CLI commands

akronim26:
# Summary 
- This is a draft PR. It merges the functioning of `mopro init` and `mopro build` into `mopro create`.

# What I have done
- I have implemented the core logic and command flow according to my current understanding of the requirements.

# What is left
- I am currently unable to fully decode or validate the testing process for this feature. I’m unsure where to start with the tests or if there are any issues in my code that need fixing.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -56,7 +56,7 @@ mopro init
 mopro build
 ```
 
-or 
+or
 
 ```sh
 mopro build --auto-update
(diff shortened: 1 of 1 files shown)


akronim26:
@team, please review!

vivianjeng (commented):
hey I think the DX looks good
and it works like what we want

but can you rename this `mopro create` with another name
e.g. `mopro full-create` or other better name

I think there is some conflict between previous `mopro create` and this command
If I want to build for 2 templates with the same rust code
it will create 2 rust projects
e.g. previously I can use
```sh
mopro init
mopro build
mopro create --framework ios
mopro create --framework android
```
but the current `mopro create` doesn't allow me to create the second template

I would like to merge this PR but I hope it can be other command for now
and in the future we can figure out a better flow
then we can convert this command to `mopro create`

akronim26:
The command `create` has been renamed to `construct` for clarity. Meanwhile, `create` still exists temporarily and behaves as it did before.

PR #654, opened 2026-01-13
refactor: avoid extra allocations in circom G2 conversion

VolodymyrBg:
This change simplifies the G2 -> CircomProverG2 conversion by parsing only the first two string coordinates of each limb directly into [BigUint; 2], instead of first collecting all entries into Vec<BigUint> and then cloning the first two elements. This removes three intermediate vector allocations and several unnecessary clones while preserving the existing behavior and FFI-facing API, including the current panic semantics on invalid input or insufficient coordinates.


vivianjeng:
thank you for the contribution

PR #705, opened 2026-05-04
Adopt noir-rs v1.0.0-beta.19 (barretenberg-rs migration)

### LINKED ISSUE #704: Adopt noir-rs v1.0.0-beta.19 (barretenberg-rs migration) across mopro
## Summary

[`zkmopro/noir-rs#37`](https://github.com/zkmopro/noir-rs/pull/37) upgraded noir-rs to `v1.0.0-beta.19` and moved from the custom `bb/` crate to the official [`barretenberg-rs`](https://crates.io/crates/barretenberg-rs). The PR explicitly defers mopro-side updates. This issue tracks them.

## Tasks

- [ ] **Upgrade the noir-rs pin** in [`cli/src/init/noir.rs`](https://github.com/zkmopro/mopro/blob/main/cli/src/init/noir.rs) — bump off `v1.0.0-beta.8-3` and drop the removed `android-compat` feature.
- [ ] **Regenerate noir test fixtures** (`noir_multiplier2.json` / `.srs` / `.vk`) under `nargo 1.0.0-beta.19` — ACIR format, SRS sizing, and proof layout all changed.
- [ ] **Update noir-related app templates** (iOS `ContentView.swift`, Android `NoirComponent.kt`, Flutter, RN) and verify a fresh `mopro init --adapter noir` builds and runs end-to-end on device.
- [ ] **Update docs** — [`docs/docs/adapters/noir.md`](https://github.com/zkmopro/mopro/blob/main/docs/docs/adapters/noir.md) (version strings, drop `x86_64-apple-ios`, drop `aztec-packages` reference) and the May 2025 noir integration blog post.
- [ ] **Update CI** — bump nargo to `1.0.0-beta.19` in `.github/workflows/build-and-test.yml` and drop `x86_64-apple-ios` from the matrix.

## References

- noir-rs PR: https://github.com/zkmopro/noir-rs/pull/37
- noir-lang release: https://github.com/noir-lang/noir/releases/tag/v1.0.0-beta.19
- Related: #422, #551

moven0831:
Closes #704. The mopro side of the noir-rs beta.19 migration ([zkmopro/noir-rs#37](https://github.com/zkmopro/noir-rs/pull/37)).

## What changes

- **CLI pin** — `cli/src/init/noir.rs`: noir-rs → `tag = "v1.0.0-beta.19"`, and drop the obsolete `android-compat` feature so fresh projects build out of the box.
- **Generic Android build knobs** — `mopro-ffi/src/app_config/android.rs`: `AndroidBindingsParams` is now a real struct (`arch_overrides`, `min_sdk_version`, `relink_with_ndk_for_bindgen`) instead of `()`. mopro-ffi stays adapter-free — it just applies whatever per-arch env/RUSTFLAGS, min SDK, and separate bindgen build the caller passes.
- **Android `dlopen` fix** — `cli/src/build/android_noir.rs` (new): all noir/barretenberg knowledge lives here, driven from `cli/src/build.rs` only when the Noir adapter is selected (mirrors `IosBindingsParams`). It fetches the correct barretenberg prebuilt and links it with Zig (root cause below).
- **Test fixtures** — regenerated `noir_multiplier2.{json,srs,vk}`, plus a new `noir_multiplier2_poseidon.vk`. beta.19 bakes the oracle hash into the VK, so the two Poseidon `_with_vk` tests read the new file; the Keccak tests and the sample apps keep using `noir_multiplier2.vk`.
- **Docs** — adapter doc, rust-setup, and the noir blog post now point at the `barretenberg-rs` crate on crates.io and drop `x86_64-apple-ios`.
- **CI** — `build-and-test.yml`: install Zig for the noir leg, raise the Android emulator to API 30, and give the Android job more timeout headroom (slower proving + the extra bindgen build).

## Android `dlopen` fix (root cause for previous CI failure)

`barretenberg-rs`'s `build.rs` picks its prebuilt by substring-matching the target triple, with the `linux` arm **before** `android`:

```rust
t if t.contains("x86_64") && t.contains("linux")   => "amd64-linux",    // matches first
t if t.contains("x86_64") && t.contains("android") => "x86_64-android", // never reached
```

`x86_64-linux-android` contains `linux`, so it downloads the glibc / libstdc++ build and links it into the Android `.so`. Android has neither, so `dlopen` fails at runtime (`__libc_single_threaded`, `std::__cxx11::*`). It's fixed upstream ([PR #20759](https://github.com/AztecProtocol/aztec-packages/pull/20759)) but that commit is unpublished, so a version bump can't pull it in.

`barretenberg-rs` honors `BB_LIB_DIR` ahead of its own download, so the Noir Android path (`cli/src/build/android_noir.rs`) now:

1. Reads the pinned `barretenberg-rs` version from the project's `Cargo.lock` — and bails loudly if it can't resolve it, rather than silently linking the wrong prebuilt.
2. Pre-fetches the matching `barretenberg-static-<arch>-android.tar.gz` and points `BB_LIB_DIR` at it (cached per version under `build/`).
3. Links the shipped lib with Zig — the prebuilt's libc++ is `std::__1`, and the NDK's `std::__ndk1` is ABI-incompatible.
4. Runs a second NDK-linked build to extract uniffi metadata, because Zig strips the `.symtab` bindgen reads.

This is a no-op for adapters that don't use barretenberg-rs. It needs Zig and the NDK on `PATH`; the builder errors clearly if either is missing.

**Min API 30**: the correct prebuilt imports `__tls_get_addr` (API 29) and `getifaddrs` (API 24), above cargo-ndk's default of 21. The builder passes `--platform 30` (noir-only) and CI runs the emulator on API 30.

## Test plan

- [x] Fresh `mopro init --adapter noir` against this CLI pins beta.19 with no `android-compat` warning.
- [x] Generated project: `cargo test --all --all-features` → 19/19 noir tests pass, all four `_with_vk` variants included.
- [x] `cargo check --workspace`; `cargo test -p mopro-cli` (covers the `barretenberg-rs` version parser).
- [x] Local `x86_64-linux-android` build: the shipped `.so` has no undefined `__libc_single_threaded` / `__cxx11` symbols, and `BB_LIB_DIR` was used.
- [x] Android instrumentation (`noirButtonClick`) and iOS simulator E2E — handled by CI.

## Out of scope (tracked under #704)

- Flutter and React Native template bumps live in their own repos.
- `docs/docs/performance.md` benchmarks still cite beta.3; re-running them is separate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Per-architecture Android build/config support and improved Android binding generation using published backend prebuilts.

* **Bug Fixes**
  * Increased emulator/test timeouts for proving and verification to reduce flakes.
  * Bumped Android emulator API level and raised Android minimum SDK target to 30 for builds.

* **Documentation**
  * Updated Noir integration to v1.0.0-beta.19, clarified dependency sourcing and platform support (x86_64 iOS simulator removed).
  * Regenerated Noir test vectors and updated verification-key tests.


vivianjeng:
@coderabbitai review

mopro-ffi/src/app_config/android.rs
```diff
@@ -88,25 +109,40 @@ impl PlatformBuilder for AndroidPlatform {
...
     mode: Mode,
-) -> anyhow::Result<PathBuf> {
+) -> anyhow::Result<(PathBuf, bool)> {
     let arch_str = arch.as_str();
     install_arch(arch_str.to_string());
     let cpp_lib_dest = bindings_out.join("jniLibs");
 
+    // barretenberg-rs's build.rs matches "linux" before "android" and fetches a
+    // glibc prebuilt Android can't dlopen; BB_LIB_DIR overrides it with the right one.
+    let bb_lib_dir = setup_barretenberg_android_lib(arch, project_dir, build_dir)?;
```
vivianjeng: In this version of `mopro-ffi`, I hope it can be adpater-free
it should not be built for any adapter (e.g. circom or noir)
if there is adapter related change, it should be in cli `cli/src/template/init/build.rs`

mopro-ffi/src/app_config/android.rs
```diff
@@ -215,3 +532,67 @@ fn reformat_kotlin_package(
...
+[[package]]
+name = "barretenberg-rs"
+version = "4.3.0"
+"#;
+        assert_eq!(
+            parse_barretenberg_rs_version(lock).as_deref(),
+            Some("4.3.0")
+        );
+    }
+}
```
vivianjeng: I don't think these tests make sense
it just mock the lock file and test the function and it will always true even the real version changes

vivianjeng (changes requested):

scripts/regen-noir-fixtures.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
vivianjeng: can we remove all the `regen-noir-fixtures` for now?
maybe after refactor we can figure out better way to compile and reuse the circuits

mopro-ffi/src/app_config/android.rs
moven0831: Agreed that `mopro-ffi` should stay adapter-free. I looked into hosting this in the generated `cli/src/template/init/build.rs` specifically, but a build script there structurally can't carry the core mechanics:

- **`BB_LIB_DIR`** has to be in the `cargo ndk` process env *before* it starts — `barretenberg-rs`'s own `build.rs` runs (and picks its prebuilt) *before* the user crate's `build.rs`, and a build script can't set env for another crate's build script.
- **The Zig linker** can't be selected from a build script — there's no `cargo::rustc-linker` directive; only `RUSTFLAGS -Clinker` / `CARGO_TARGET_<triple>_LINKER` / `.cargo/config.toml` can, and cargo-ndk overrides the latter two.
- **`--platform`** and the **second NDK-linked bindgen build** are `cargo ndk` orchestration, not something a templated project file can express.

So I kept the intent but moved it to the CLI build driver, mirroring the existing `IosBindingsParams { using_noir }` pattern:

- `mopro-ffi/src/app_config/android.rs` is adapter-free again (323 lines, was 217 on `main`). `AndroidBindingsParams` is now a generic struct — `arch_overrides` (per-arch env + RUSTFLAGS), `min_sdk_version`, `relink_with_ndk_for_bindgen` — with no barretenberg/noir/Aztec strings.
- All the noir/barretenberg logic (prebuilt download, version parsing, Zig wrapper, NDK discovery) moved to a new `cli/src/build/android_noir.rs`, called from `cli/src/build.rs` only when the Noir adapter is selected.

Since `mopro build` drives the Android build in-process via the CLI, this needs no `mopro-ffi` release — the generated project still compiles against the published `=0.3.6` untouched. Side effect: non-noir Android builds no longer inherit the `--platform 30` bump this PR had applied to every adapter.

Does this direction work for you?

mopro-ffi/src/app_config/android.rs
```diff
@@ -140,9 +176,290 @@ fn build_for_arch(
...
+        build_cmd.env("BB_LIB_DIR", bb_lib_dir);
+    }
+    build_cmd
+        .env("CARGO_BUILD_TARGET_DIR", &bindgen_target)
+        .env("CARGO_BUILD_TARGET", arch_str)
+        .env("CARGO_NDK_OUTPUT_PATH", bindgen_target.join("jniLibs"))
+        .spawn()
+        .expect("Failed to spawn cargo build for bindgen lib")
+        .wait()
+        .expect("cargo build (bindgen lib) errored");
```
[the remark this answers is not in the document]
moven0831: Done in eb8a6df. Both `cargo ndk` invocations now capture the exit status and `bail!` with the target arch on failure — the main `build_for_arch` build and this bindgen build. Heads-up that the code moved: `mopro-ffi`'s Android builder is now adapter-free, so this second NDK-linked build lives behind a generic `relink_with_ndk_for_bindgen` flag rather than being noir-specific.

docs/docs/adapters/noir.md
```diff
@@ -20,17 +20,16 @@ To get started, follow the [Rust Setup Guide](/setup/rust-setup.md) and ensure t
...
 ```
 
 :::info
 The Noir adapter depends on [`zkmopro/noir-rs`](https://github.com/zkmopro/noir-rs). Please checkout the usage and the version here.
 
-**Current version**: Supports Noir `1.0.0-beta.8-3` with bb `1.0.0-nightly.20250723` with updated dependencies and enhanced functionality.
+**Current version**: Supports Noir `v1.0.0-beta.19` via the official [`barretenberg-rs`](https://crates.io/crates/barretenberg-rs) crate (`=4.2.0-aztecnr-rc.2`).
 
-**Backend**: The adapter uses the Barretenberg backend, which is automatically downloaded from the released GitHub artifacts at [zkmopro/aztec-packages](https://github.com/zkmopro/aztec-packages/releases).
+**Backend**: The adapter uses the Barretenberg backend via the official [`barretenberg-rs`](https://crates.io/crates/barretenberg-rs) crate from crates.io. Prebuilt FFI binaries are fetched automatically at build time — no separate download script required.
```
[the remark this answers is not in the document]
moven0831: Verified — keeping as written. `noir-rs` `v1.0.0-beta.19` declares the dependency as `barretenberg-rs = { version = "=4.2.0-aztecnr-rc.2", default-features = false, features = ["ffi"] }` in its `noir/Cargo.toml`, i.e. an exact crates.io version constraint, not just an Aztec tooling tag. So "via the official `barretenberg-rs` crate (`=4.2.0-aztecnr-rc.2`)" is accurate. Happy to add a sentence distinguishing the crate version from the Barretenberg binary alignment if you'd still like that called out.

scripts/regen-noir-fixtures.sh
moven0831: Done in bf25096 — removed `scripts/regen-noir-fixtures.sh`, the helper crate, and the now-orphaned `test-vectors/noir/multiplier2` circuit source. The committed `.json` / `.srs` / `.vk` fixtures the tests consume are untouched. Agreed on figuring out a better circuit compile/reuse story after the refactor.

mopro-ffi/src/app_config/android.rs
moven0831: Good point — the two version-extraction tests just re-derived their mocked input, so I removed those in bf25096. (The parser moved to `cli/src/build/android_noir.rs` with the adapter-free refactor.) I kept the single case that guards a real pitfall: not mistaking a `[[package]]` that *lists* `barretenberg-rs` as a dependency for the actual `barretenberg-rs` package entry. Happy to drop that one too if you'd rather have no unit coverage there.

scripts/regen-noir-fixtures.sh
```diff
@@ -0,0 +1,47 @@
...
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+CIRCUIT_DIR="$ROOT/test-vectors/noir/multiplier2"
+TARGET_DIR="$CIRCUIT_DIR/target"
+TEMPLATE_VEC="$ROOT/cli/src/template/init/test-vectors/noir"
+TOPLEVEL_VEC="$ROOT/test-vectors/noir"
+
+echo "[1/4] Compiling multiplier2 with $(nargo --version | head -1)"
+( cd "$CIRCUIT_DIR" && nargo compile )
```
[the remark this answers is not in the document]
moven0831: Moot now — the `regen-noir-fixtures` script was removed in bf25096 per the maintainer request, to be revisited after a circuit compile/reuse refactor.

PR #725, opened 2026-07-29
fix: replace dead ci-keys.zkmopro.org links

### LINKED ISSUE #724: [BUG] rapidsnark.zkmopro.org no longer resolves
## Symptom

Building a project that uses `ProofLib::Rapidsnark` fails:

```
curl: (6) Could not resolve host: rapidsnark.zkmopro.org
```

## Cause

`rust-rapidsnark`'s build script fetches prebuilt static libs from `https://rapidsnark.zkmopro.org/$TARGET.zip` ([`crates/download_rapidsnark.sh`](https://github.com/zkmopro/rust-rapidsnark/blob/main/crates/download_rapidsnark.sh)), and `crates/build.rs` panics when the download fails.

That host is gone:

```
$ dig @8.8.8.8 rapidsnark.zkmopro.org
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN

$ dig @8.8.8.8 zkmopro.org +short
172.67.142.13
104.21.63.10
```

Same result at 1.1.1.1. The zone still resolves, so a single subdomain record is missing. An R2 custom domain that got removed would explain it, since deleting the binding also deletes the DNS record.

`ci-keys.zkmopro.org` is NXDOMAIN too, and our docs tell users to download zkeys from it: `docs/docs/setup/rust-setup.md:174-175`, `docs/docs/sdk/react-native.md:120`, `docs/blog/2025-03-27-ethtaipei-workshop.md:239-240,318`.

No mirror of the rapidsnark zips exists anywhere. `zkmopro/rapidsnark` has no tags and no releases, the Wayback Machine archived only a 404 of the root, and org-wide code search finds the hostname in exactly one file.

## Who this hits

Anyone who enables the `rapidsnark` feature on `circom-prover`, which `docs/docs/adapters/circom.md:304-314` tells you to do for `ProofLib::Rapidsnark`.

Default scaffolded projects build fine, because `cli/src/init/circom.rs:13-17` writes `circom-prover = "0.1"` with no features, so `rust-rapidsnark` is never compiled. That is why mopro's CI missed it. `zkmopro/rust-rapidsnark` does cross-build every affected target in its own CI and would have caught this, but it has not run since September 2025. Same gap, separate bug: with the feature off, `ProofLib::Rapidsnark` reaches `panic!("Unsupported proof library")` at runtime, because `circom-prover` 0.1.4 gates the match arms but not the enum variant.

## Proposed fix

Download from `iden3/rapidsnark` v0.0.8 release assets instead. Checked before proposing it:

- v0.0.8 publishes a zip for every target `rust-rapidsnark` supports except `x86_64-apple-ios` (see below), each containing the four archives `build.rs` links (`librapidsnark.a`, `libfr.a`, `libfq.a`, `libgmp.a`).
- Its libs export the two symbols `rust-rapidsnark` declares, `_groth16_prover_zkey_file` and `_groth16_verify`.
- `rapidsnark-iOS-Simulator-v0.0.8.zip` carries universal `librapidsnark.a` and `libgmp.a`, so it serves `aarch64-apple-ios-sim`.
- It does not cover `x86_64-apple-ios`. `libfr.a` and `libfq.a` in that asset are built arm64 only, in every upstream release, and the x86_64 slice of `librapidsnark.a` needs 29 symbols from them (hand-written field-arithmetic assembly). `x86_64-apple-ios` is therefore dropped from `rust-rapidsnark`'s supported targets and now fails with a message pointing at `RAPIDSNARK_LIB_DIR`. mopro lists it as an iOS arch (`mopro-ffi/src/app_config/constants.rs:129`), so Intel Macs lose simulator rapidsnark and the docs need a note.
- `zkmopro/rapidsnark` is 2 commits ahead of upstream (a makefile tweak and a gmp script fix) and 56 behind, so we lose nothing by switching.

Two wrinkles. The iOS and iOS-Simulator zips keep the `.a` files flat while everything else nests them under `lib/`, so the script normalizes both into `$OUT_DIR/rapidsnark/<target>/`. And rustc rejects universal archives when bundling a static library, because the arm64e slice of the iOS `libgmp.a` and the arm64 slice of the simulator `librapidsnark.a` are not valid ar archives on their own, failing with `Unsupported archive identifier`. Each Apple archive is thinned with `lipo` to the slice being built for.

While in that script, add `RAPIDSNARK_LIB_DIR` and `RAPIDSNARK_DOWNLOAD_BASE_URL` overrides so a build can point at local or mirrored libs. We already do this with `BB_LIB_DIR` in `cli/src/build/android_noir.rs`, and `rust-gnark` ships `RUST_GNARK_PREBUILT_URL`.

Publishing `rust-rapidsnark` 0.1.4 unblocks fresh builds on its own, since `circom-prover` asks for caret `"0.1"`. Anyone with an existing lockfile needs `cargo update -p rust-rapidsnark`.

## Other build-time downloads, audited

Healthy: barretenberg (Aztec release `v4.2.0-aztecnr-rc.2`, both Android assets present), `rust-gnark` prebuilts, `noir_rs` (pinned to tag `v1.0.0-beta.19`), and the Kotlin/JNA/wasm-pack fetches in CI.

Working but unpinned, so exposed to the same class of failure: `witnesscalc-adapter` clones `zkmopro/witnesscalc`'s default branch on every circom build and then runs `git checkout v2.2.0`, where `v2.2.0` is a branch rather than a tag (that repo has no tags at all). The `frb` and `ubrn` template archives and the `uniffi-bindgen-react-native` clone are unpinned as well.

## Tasks

- [x] `zkmopro/rust-rapidsnark`: switch the download to `iden3/rapidsnark` v0.0.8, normalize the archive layout, thin the Apple archives, add the two env overrides, verify checksums
- [x] Publish `rust-rapidsnark` 0.1.4
- [ ] Add CI cover in this repo that builds a scaffolded circom project with the `rapidsnark` feature on
- [ ] Note the `x86_64-apple-ios` gap in the mopro circom docs
- [ ] Replace the dead `ci-keys.zkmopro.org` URLs in the docs
- [ ] Check the Cloudflare zone for other records that went missing

moven0831:
# Fix dead `ci-keys.zkmopro.org` links

- Follow-up to and close #724, alongside zkmopro/rust-rapidsnark#12.

## Problem

`ci-keys.zkmopro.org` no longer resolves (same lapsed zone as `rapidsnark.zkmopro.org`), so the Rust setup guide, React Native SDK page, and EthTaipei workshop all fail with `curl: (6) Could not resolve host`.

## Changes

Every link now points at a fixture already committed in the org, pinned to a tag or commit so it can't drift:

| File | New source |
| --- | --- |
| `multiplier2.wasm`, `multiplier2_final.zkey` | this repo @ `mopro-cli-v0.3.7` |
| keccak256 wasm + zkey | rust-rapidsnark @ `v0.1.4` (only committed copies) |
| `keccak256.json` | circuit-registry `keccak256/input.json` — verified identical to the 256 values the workshop pastes inline |

The 0.1 and 0.2 docs carried the same dead host and are updated too.

**Also:**

- **Circom adapter page** now notes that `rust-rapidsnark` 0.1.4 dropped `x86_64-apple-ios` (upstream builds `libfr.a`/`libfq.a` arm64-only for the simulator). mopro still offers that arch, so an Intel Mac would otherwise fail with no explanation.
- **`cli_template_tests`** now builds a scaffolded circom project with the `rapidsnark` feature enabled. Nothing in CI compiled `rust-rapidsnark` before — scaffolded projects leave the feature off — which is how a download endpoint could break for users while CI stayed green. Runs on both runners; the Ubuntu leg would have caught the Linux link failure in rust-rapidsnark#12.

## Summary by CodeRabbit

* **New Features**
  * Enhanced the scaffold validation to build Circom projects with the Rapidsnark feature enabled.

* **Tests**
  * Added an explicit verification that the required feature is present before building, and improved the related failure message.

* **Documentation**
  * Updated Circom, Rust, React Native, and workshop guides to use newer raw GitHub download links for circuit artifacts and test vectors.
  * Added clarification about missing prebuilt iOS covers and how to build the Intel Mac simulator using a local library path.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -132,6 +132,18 @@ jobs:
                   export CLASSPATH="$PWD/jna-5.13.0.jar"
                   cargo test --all --all-features
 
+            - name: Build with the rapidsnark feature
+              if: matrix.adapter == 'circom'
+              working-directory: ${{ runner.temp }}/mopro-example-circom
+              run: |
+                  sed -i.bak 's/^circom-prover = "0.1"$/circom-prover = { version = "0.1", features = ["rapidsnark"] }/' Cargo.toml
+                  rm Cargo.toml.bak
+                  grep -q 'features = \["rapidsnark"\]' Cargo.toml || {
+                      echo "::error::failed to enable the rapidsnark feature; the circom-prover entry in write_toml.rs may have changed"
+                      exit 1
+                  }
+                  cargo build
+
     cli_build_ios_prepare:
         needs: cli_build
         strategy:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -139,7 +139,7 @@ jobs:
                   sed -i.bak 's/^circom-prover = "0.1"$/circom-prover = { version = "0.1", features = ["rapidsnark"] }/' Cargo.toml
                   rm Cargo.toml.bak
                   grep -q 'features = \["rapidsnark"\]' Cargo.toml || {
-                      echo "::error::failed to enable the rapidsnark feature; the circom-prover entry in write_toml.rs may have changed"
+                      echo "::error::failed to enable the rapidsnark feature; the circom-prover entry in cli/src/init/circom.rs may have changed"
                       exit 1
                   }
                   cargo build
--- a/docs/blog/2025-03-27-ethtaipei-workshop.md
+++ b/docs/blog/2025-03-27-ethtaipei-workshop.md
@@ -236,8 +236,8 @@ Alternatively, you can watch this video to see how to run the app.
 This section explains how to update circuits with alternative witness generators and corresponding zkey files. We use the [Keccak256 circuit](https://github.com/zkmopro/circuit-registry/blob/main/keccak256/keccak256_256_test.circom) as a reference example here.
 
 1.  Add wasm and zkey file in the `test-vectors/circom` folder
-    -   wasm: https://ci-keys.zkmopro.org/keccak256_256_test.wasm
-    -   zkey: https://ci-keys.zkmopro.org/keccak256_256_test_final.zkey
+    -   wasm: https://raw.githubusercontent.com/zkmopro/rust-rapidsnark/v0.1.4/tests/test-vectors/keccak256_256_test.wasm
+    -   zkey: https://raw.githubusercontent.com/zkmopro/rust-rapidsnark/v0.1.4/tests/test-vectors/keccak256_256_test_final.zkey
 2.  In `src/lib.rs` file, update the circuit's witness generator function definition.
 
     ```diff
@@ -315,7 +315,7 @@ This section explains how to update circuits with alternative witness generators
     + val zkeyPath = getFilePathFromAssets("keccak256_256_test_final.zkey")
     ```
 
--   Update circuit inputs: https://ci-keys.zkmopro.org/keccak256.json
+-   Update circuit inputs: https://raw.githubusercontent.com/zkmopro/circuit-registry/aab2ee1290ecd5d2b7dff985520bd1f3ce8a9750/keccak256/input.json
 
     -   **iOS:**<br/>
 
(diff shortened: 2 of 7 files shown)


moven0831:
`rapid-snark` feature check on CI has been added. Logs as below:

https://github.com/zkmopro/mopro/actions/runs/30463110309/job/90618885679#logs

